PR #149, opened 2015-01-09
Add HEAD request test

lexmag:
Tests 7b60ae08.


edgurgel:
Shall we also add an async head request?

lexmag:
I think so, I'm planning to add an async HEAD test after merge of #151 since it contains async test refactoring.

Issue #118, opened 2014-08-21
hackney_response:read_body typespec


edgurgel:
https://github.com/benoitc/hackney/blob/master/src/hackney_client/hackney_response.erl#L299-L311

Here it says that `read_body` can return `{ok, binary(), #client{}} | {error, atom()}.`, but it can return `{error, {closed, << Acc/binary, Bin/binary >>}};`, right?

Which won't be an atom, right?

BTW, if no body is coming from the request, hackney will return an error or just an empty binary?

PR #151, opened 2015-01-09
Fix handling of bodiless (204, 304 codes) responses

lexmag:
It fixes the problem with 304 and 204 responses that don't provide `Content-Lenght` header (e.g. google API).
Reading of body halts until `{error {closed, <<>>}}` after a long timeout.

The returning tuple for this response changes to `{ok, Status, Headers}`, same as for HEAD requests 7b60ae08 (let me know if you'd like to keep current `{ok, Status, Headers, Ref}` behaviour).

The httpbin has no way not to provide `Content-Lenght` thus I'm looking for another way to test the lack of `Content-Lenght`, that I think will be an another PR.

Relates to #118. \cc @edgurgel

### DIFF
--- a/src/hackney_client/hackney.erl
+++ b/src/hackney_client/hackney.erl
@@ -86,7 +86,7 @@ connect(Transport, Host, Port, Options) ->
 
 %% @doc Assign a new controlling process <em>Pid</em> to <em>Client</em>.
 -spec controlling_process(client_ref(), pid())
-	-> ok | {error, closed | not_owner | atom()}.
+    -> ok | {error, closed | not_owner | atom()}.
 controlling_process(Ref, Pid) ->
     hackney_manager:controlling_process(Ref, Pid).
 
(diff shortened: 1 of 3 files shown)


edgurgel:
Awesome! I'm thinking on adding a nobody case on httpinb.org/status/204 and httpinb.org/status/304 . As soon as I finish it and add tests I will send a PR to httpbin

edgurgel:
This may also be related to this: https://github.com/edgurgel/httpoison/issues/33

Async requests when Content-length is 0 never receives the `done` message.

edgurgel:
One way to see this happening is running httpbin locally and doing a GET to `/status/304`.

```
$ curl -i http://localhost:8000/status/304
HTTP/1.1 304 NOT MODIFIED
Server: gunicorn/18.0
Date: Sun, 11 Jan 2015 04:57:33 GMT
Connection: close
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
```
(If you use httpbin.org it will (for some reason) have a content-length header)

The async request hangs waiting for a done message.

lexmag:
@edgurgel thanks for pointing out to hanging of async request.
I pushed fix for this case as well. I believe it should fix edgurgel/httpoison#33.

benoitc:
patch is OK for me. Hopefully all servers are following the specifications (especially http://svn.tools.ietf.org/svn/wg/httpbis/specs/rfc7231.html#status.204). I will have a quick look on it today and if no issue merge it.

edgurgel:
I know we have many faulty HTTP servers out there, but I still doubt that someone is doing 204, 304 with bodies. Hopefully nobody will have any trouble with this approach :+1:

lexmag:
Rebased.

PR #154, opened 2015-01-18
Make bodiless responses consistent with the rest of API

### LINKED ISSUE #153: 1.0.5 broke API when response has no body
When you request an endpoint that has no content, then the return value of hackney used to be

```erlang
{ok, StatusCode, RespHeaders, ClientRef}
```

whereas it is

```erlang
{ok, StatusCode, RespHeaders}
```

for 1.0.5. An example endpoint is here https://developer.github.com/v3/gists/comments/#delete-a-comment

lexmag:
Closes #153


sdepold:
:+1:

edgurgel:
:+1:

PR #156, opened 2015-01-21
Correct 'keep-alive' response header

tnt-dev:


benoitc:
Hrm nice catch! thanks for the patch :) I am pretty sure though I saw some server sending keepalive, would you mind to handle both cases in your patch?

kxepal:
@benoitc if you google about "Connection: keepalive" you'll indeed find few of these cases. However, they aren't belongs to any popular product and obliviously violates HTTP RFC definition where:

```
19.7.1.1 The Keep-Alive Header
   When the Keep-Alive connection-token has been transmitted with a
   request or a response, a Keep-Alive header field MAY also be
   included. The Keep-Alive header field takes the following form:

          Keep-Alive-header = "Keep-Alive" ":" 0# keepalive-param

          keepalive-param = param-name "=" value
```

benoitc:
okaye. so let's merge that now. Thanks!

PR #167, opened 2015-02-23
add reuse_sessions to all cases

cstar:
Could this help?


cstar:
pointless (it's the default value — https://github.com/erlang/otp/blob/OTP-17.3.4/lib/ssl/src/ssl.erl#L646) , closing it.

PR #175, opened 2015-03-14
removing extra CRLF at the start of buffer before passing it to `parse_headers`

arashbm:
a `<<"\r\b">>` at the beginning of the buffer passed to `parse_headers/1` prevented parser from correctly parsing headers. It is correctly deleted when handling responses. (see line 194 of `src/hackney_lib/hackney_http.erl`)

### DIFF
--- a/src/hackney_lib/hackney_http.erl
+++ b/src/hackney_lib/hackney_http.erl
@@ -247,7 +247,7 @@ parse_uri_path(<< C, Rest/bits >>, St, Method, Acc) ->
         _ -> parse_uri_path(Rest, St, Method, << Acc/binary, C >>)
     end.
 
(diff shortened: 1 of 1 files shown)


benoitc:
Patch looks good. Do you have any code that could help to test it? Eventually some eunit that we could add.

arashbm:
I'll write some tests. Since submitting this patch I [fixed another bug](https://github.com/arashbm/hackney/compare/57f48ef9e...06981d066) in parsing requests on a presistant connection. I will add a bunch of tests for both these cases in `test/hackney_http_tests.erl`.

benoitc:
ca you send me a PR with the changes above. Maybe we could add your patches in the coming release today?

PR #198, opened 2015-06-23
move to rebar3

tsloughter:
With this you could also take over the hex.pm package and publish it with `rebar3 hex publish` or `rebar3 hex cut` http://www.rebar3.org/v3.0/docs/using-available-plugins#hex-package-management


benoitc:
Thanks! I will take care about that in the morning and make sure that the teating is working again.

edgurgel:
Do we need to change travis to use rebar3 to get-deps, build and test. Right now it's using good old `rebar` executable. https://travis-ci.org/benoitc/hackney/jobs/68348081

benoitc:
+1 for the change

edgurgel:
https://github.com/benoitc/hackney/pull/202

PR #202, opened 2015-06-27
Change .travis.yml to use rebar3

edgurgel:
This should be enough to make Travis CI use rebar3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,5 @@ before_install:
   - sudo pip install -q httpbin
   - sudo pip install -q gunicorn
 before_script: gunicorn httpbin:app&
+install: ./rebar3 update
+script: make test


edgurgel:
For some reason R15 and R15B1 are failing to fetch the dependencies:

```
===> Verifying dependencies...
===> Fetching ssl_verify_hostname ({pkg,<<"ssl_verify_hostname">>,
                                           <<"1.0.5">>})
===> Failed to fetch and copy dep: {pkg,<<"ssl_verify_hostname">>,<<"1.0.5">>}
```

benoitc:
imo we should just switch to R18 for 2.0 at least R17. thoughts?

edgurgel:
@benoitc, seems legit 17+ for 2.0. Any plans to release a working branch? I was thinking on writing unit tests for most modules but I don't want to start when the code may change drastically :D

@tsloughter, is rebar3 supposed to support somehow R15? For some reason it's breaking on Travis during download. 

I'm ok to drop support to R15 if @benoitc agrees.

tsloughter:
Actually yes, we test against R15B03-1. Technically we would be removing it now that 18 is out (we plan to only support current OTP and 2 previous majors), but I don't see the point right now.

benoitc:
@edgurgel I will merge tomorrow current work from the branch [hackney_socks](https://github.com/benoitc/hackney/tree/hackney/hackney_socks) and latest master tomorrow. In the process I will probably make some changes less drastic.

edgurgel:
I just noticed that the hackney integration tests are not running with rebar3...

benoitc:
@edgurgel can you rebase against the latest version of hackney to check the tests are running?

benoitc:
fixed in latest master. Thanks!

edgurgel:
Too fast :)

edgurgel:
I wonder what's the best way to run the integration tests using rebar3 cause they are not running :/ We went from 159 to 138 tests after this change

jlouis:
create an empty `hackney_integration.erl` module :) rebar3 only runs a test module if it corresponds to a real module IIRC.

edgurgel:
Ok! Apparently running `./rebar3 eunit --dir="test"` does the trick

PR #204, opened 2015-07-07
Use erlang:byte_size/1 for size of binaries

puzza007:
Not a bug but we only need the size of binaries, not binaries and tuples.  Also removed a redundant call to `iolist_to_binary`


benoitc:
Thanks!

PR #208, opened 2015-07-09
adding Equifax Secure Certificate Authority

odo:
Apple removed the Equifax root certificate so GeoTrust and RapidSSL won't work on OS X 10.10.4 (14E46) anymore.
This patch adds this certificate.

http://www.infoworld.com/article/2911209/mac-os-x/yosemite-10103-breaks-some-applications-and-https-sites.html
https://www.geotrust.com/resources/root-certificates/


benoitc:
new chain in 99434ee79076c014945e0785101c0d229515fb78 include it. Thanks!

PR #209, opened 2015-07-09
Fix typo: ssl_opttions -> ssl_options

Frost:
This typo appeared in two places in src/hackney_client/hackney.erl, and does not
seem to be related to anything except themselves.


benoitc:
thanks!

PR #210, opened 2015-07-12
Update README.md

sarah:
Host on hackney:connect should not include protocol (given by the transport). If you put the protocol in, you get `{error, nxdomain}` back from gen_tcp:connect().  Also for an SSL request, `hackney_ssl_transport` is the Transport. see: [{hackney_ssl_transport, 443} and {hackney_tcp_transport, 80}](https://github.com/benoitc/hackney/blob/master/src/hackney_connect/hackney_connect.erl#L52-L53)

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,8 +174,8 @@ couple of requests.
 
 ```erlang
 
-Transport = hackney_tcp_transport,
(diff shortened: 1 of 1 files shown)


benoitc:
Thanks!

PR #221, opened 2015-08-01
fix typo in readme

mcfiredrill:


benoitc:
Thanks for the patch. I applied Iit by porting the change to the `overview.doc` file which is used to generate the README.

PR #222, opened 2015-08-07
connect_time metric is a histogram

mweibel:


benoitc:
Thanks! However the readme is generated from the overview. Would you mind to edit it directly. I can do the change myself anyway, just let me know :)

mweibel:
oh, ok :D have a look at #227

PR #236, opened 2015-08-27
Fix parsing URLs with query string after the domain name

tnt-dev:
At this moment `hackney` can't handle URLs like `www.example.com?key=value` because it hasn't slash after domain.


benoitc:
Thanks!

PR #237, opened 2015-08-28
fix redirect to relative url

tnt-dev:
- Don't override `Host` header
- Merge redirect url with current path


benoitc:
Thanks for the patch! However, the integration tests don't pass not sure what's wrong yet, can you fix them too?

tnt-dev:
Sorry, It's really my fault. PR updated.

benoitc:
Thanks. Merged. Can you add a test for it anyway?

PR #244, opened 2015-09-13
Update README.md

deadtrickster:
use `./rebar3 shell` command to `run a shell with project apps and deps in path.` (old -pa stuff doesn't work anymore)

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,14 @@ To start in the console run:
 ```erlang-repl
 
 $ ./rebar3 shell
+
+```
+
+It is suggested that you install rebar3 user-wide as described [here](http://blog.erlware.org/rebar3-features-part-1-local-install-and-upgrade/).
+This fixes zsh (and maybe other shells) escript-related bugs. Also this should speed things up.
+
+```erlang-repl
+
 1>> hackney:start().
 ok
 ```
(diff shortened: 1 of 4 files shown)


benoitc:
Thanks for the patch. Hoever the change should be made in the overview file instead. The readme is generated from it. Can you do the changes?

Also I wonder if we shouldn't add a note for zsh users (on which the script is buggy) yto use the command line `erl -pa _build/default/*/ebin` instead. Thoughts?

deadtrickster:
I just tried on zsh 5.0.5 - works perfectly!

benoitc:
hrm will retry on my mac. when you exit here the text is scrambled

benoitc:
@deadtrickster about the issue: 
https://github.com/rebar/rebar3/issues/644#issuecomment-140117818

What about documenting both? Can you do that change?

deadtrickster:
How about these lines: "It is suggested that you install rebar3 user-wide as described here http://blog.erlware.org/rebar3-features-part-1-local-install-and-upgrade/. This fixes zsh (and maybe other shells) escript-related bugs. Also this should speed things up."

benoitc:
:+1:

deadtrickster:
done

benoitc:
Thanks!

PR #245, opened 2015-09-29
fix default name value of mp file header

falood:
Is this a bug?

http://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.2


benoitc:
good catch!

Issue #246, opened 2015-10-02
Unwanted encoding of "$" in URL path


skirino:
Similar to https://github.com/benoitc/hackney/issues/176 but this time is about `$`.

I'm trying to access to a URL whose path contains `$`, e.g.
```
/jenkins/$stapler/bound/e7b78ef7-b201-4cc2-bae6-3cf2d41f6ed4/render
```
hackney converts `$` to `%24` before sending request and results in 404 NotFound error.

For reference:
- http://tools.ietf.org/html/rfc3986#section-2.2
- http://stackoverflow.com/questions/20050114/what-does-dollar-sign-do-in-url
- https://github.com/elixir-lang/elixir/blob/v1.0.5/lib/elixir/lib/uri.ex#L180

deadtrickster:
excerpt from rfc: 
```
URI producing applications should percent-encode data octets that
   correspond to characters in the reserved set unless these characters
   are specifically allowed by the URI scheme to represent data in that
   component.  If a reserved character is found in a URI component and
   no delimiting role is known for that character, then it must be
   interpreted as representing the data octet corresponding to that
   character's encoding in US-ASCII.
```
I'm not sure $ has any special meaning in http context. looks more like url parsing/routing failure on the server-side

skirino:
@deadtrickster Thank you for your comment.
I believe that [Sean B. Durkin's answer in stackoverflow](http://stackoverflow.com/questions/20050114/what-does-dollar-sign-do-in-url?answertab=votes#tab-top) is correct, i.e. `$` has no special meaning in the URI syntax.
Actually [http_uri:encode/1](http://erlang.org/doc/man/http_uri.html#encode-1) does not convert `$`:
```erl
1> http_uri:encode("$").
"$"
```

benoitc:
The hackney URI parser is based on the chromium one and http_uri does less.

Reading the [RFC 3986](http://tools.ietf.org/html/rfc3986#section-2.2) , reserved characters should be percent encoded `unless these characters are specifically allowed by the URI scheme to represent data in that  component`.

Passing a list of reserved chars to the request may be a solution. Thoughts?

dtykocki:
@benoitc -- I have a potential fix to this issue on a local branch. If y'all decide not to go down the reserved chars path, I'll submit a PR.

skirino:
@benoitc 
Thank you for your response!
[...]

Issue #176, opened 2015-03-16
won't encode special character in http get url path


linbo:
Hi,

I have an restful server with an url like this:

```
http://host:port/id/name:107/?type=FS
```

Since there is an special character colon in the path, I've checked code,  find those codes in hackney.erl

```
request(Method, #hackney_url{}=URL0, Headers, Body, Options0) ->
    %% normalize the url encoding
    URL = hackney_url:normalize(URL0),
```

Actually, I won't wan't encode colon in URL, but maybe in query parameters, but the code seems encode all the URL.

If I use httpc:get, looks working fine.

benoitc:
mmm Is this a valid syntax though? Hackney is reusing the algorithm you can find in chrome. I will check the spec anyway.

PR #249, opened 2015-10-09
Fix handling of question mark in URL

tnt-dev:
Similar to #176 and, maybe, #246.


benoitc:
thanks! I need to check what the spec says exactly about that. There will be a new release of Hackney next week and will make sure to have it if it follows the spec.

silviucpp:
Hello,

The new pool management is part of this release ?

Silviu

benoitc:
yes :)

benoitc:
Closing as outdated. This PR is from 2015 and the URL encoding logic has been significantly refactored since then. The file path has also changed from `src/http/hackney_url.erl` to `src/hackney_url.erl`. Thank you for the contribution!

PR #252, opened 2015-10-24
Add .PHONY to Makefile

dtykocki:
`make` will only run if there isn't a
file or directory with the same name in the current
working directory. This adds .PHONY to the Makefile
so we can run `make test`.


benoitc:
thanks!

Issue #197, opened 2015-06-23
Multipart EOF don't have \r\n at the end


eminarcissus:
I've used HTTPosion to do some multiparts uploads with hackney but found the server didn't recognized the request, and I've done quite a lot of tests in modifying src/lib/hackney_multiparts.erl to find out which part is the real culprit, finally figure out looks like the request issues from requests(python) have a \r\n at the end of request but not for the hackney lib.
here is my modification:

line 137:
mp_eof(Boundary) ->
    <<"--",  Boundary/binary, "--">>.

to
mp_eof(Boundary) ->
    <<"--",  Boundary/binary, "--\r\n">>.

And after the change everything works like a charm, don't know which one did really follow the rfc sheet, but If I'm remembered well every sentence communication via http should have a \r\n at the end, so maybe this is a bug over here.

Also for this function
mp_file_header({file, Path, ExtraHeaders}, Boundary) ->
    FName = hackney_bstr:to_binary(filename:basename(Path)),
    Disposition = {<<"form-data">>,
                   [{<<"name">>, <<"file">>},
                    {<<"filename">>, <<"\"", FName/binary, "\"">>}]},
    mp_file_header({file, Path, Disposition, ExtraHeaders}, Boundary);
mp_file_header({file, Path, {Disposition, Params}, ExtraHeaders}, Boundary) ->
    CType = hackney_mimetypes:filename(Path),
    Len = filelib:file_size(Path),
    ExtraHeaders0 = lists:map(fun ({K, V}) -> {hackney_bstr:to_lower(K), V} end, ExtraHeaders),
    Headers = mp_filter_header([{<<"content-type">>, CType},
                                {<<"content-length">>, Len}],
                               [{<<"content-disposition">>, Disposition, Params} | ExtraHeaders0]),
    BinHeader = mp_header(Headers, Boundary),
    {BinHeader, Len}.

When sending Disposiion,ExtraHeaders, Disposition kv values didn't add quote " into the disposition,although it didn't make difference to the server, but it also is a point where it differs from what I get with python.requests.

from hackney
Content-Disposition: form-data; name=file; filename=avatar.jpg

from python
Content-Disposition: form-data; name="file"; filename="avatar.jpg"

so my temporary solution is add a quote into the value of name and filename temporary to make it works as expected.

Hope those can be merged and fixed soon, really appreciate your works on this lib :)

benoitc:
[...]

PR #258, opened 2015-11-20
Use cowboy to test multipart/form-data

jlouis:
This PR introduces a small test case for the multipart upload. Currently it fails with a 500 error inside cowboy because hackney is doing things which is against the RFC. The PR is part of fixing #197 since it provides the counter-example for the current operation.

I envision we can take this PR and a fix together so we can avoid regressions on this area later on.

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -21,7 +21,8 @@ Rebar2Deps = [{idna, ".*",
 
               {ssl_verify_hostname, ".*",
                {git, "https://github.com/deadtrickster/ssl_verify_hostname.erl",
-                {tag, "1.0.5"}}}],
+                {tag, "1.0.5"}}},
+              {cowboy, ".*", {git, "https://github.com/ninenines/cowboy.git", {tag, "1.0.4"}}}],
 
 case IsRebar3 of
     true ->
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,6 @@ otp_release:
   - R16B02
   - R16B01
   - R16B
-  - R15B03
-  - R15B02
-  - R15B01
-  - R15B
 before_install:
   - sudo pip install -q httpbin
   - sudo pip install -q gunicorn
--- a/rebar.config
+++ b/rebar.config
@@ -34,5 +34,9 @@
                            {packages, false},
                            {subpackages, true},
                            {top_level_readme,
-                            {"./README.md", "http://github.com/benoitc/hackney"}}]}
-             ]}]}.
+                            {"./README.md", "http://github.com/benoitc/hackney"}}]}]},
+             {test, [
+               {deps, [
+                  {cowboy, {git, "https://github.com/ninenines/cowboy.git", {tag, "1.0.4"}}}
+               ]} ]}
+           ]}.
--- a/src/http/hackney_multipart.erl
+++ b/src/http/hackney_multipart.erl
@@ -135,7 +135,7 @@ mp_header(Headers, Boundary) ->
 
 %% @doc return the boundary ennding a multipart
 mp_eof(Boundary) ->
-    <<"--",  Boundary/binary, "--">>.
+    <<"--",  Boundary/binary, "--\r\n">>.
 
 %% @doc create a part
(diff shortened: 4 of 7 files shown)


jlouis:
Analysis of the failing test case: the test fails because it uses rebar 2.5.x and it doesn't understand test profiles. Hence it doesn't understand the necessity of downloading cowboy to the system in order to make it work. Hopefully I can solve this by pushing an explicit test_deps section in rebar.config and make this go through locally. 

I think the other failing test case is a fallout from the teardown not being run correctly.

jlouis:
Progress: Working down to R16B1 now. Current culprit is that ensure_all_started is not present there. It should be a fairly easy fix if you still want to track older releases.

jlouis:
Still being trolled by R16B1 and earlier :)

jlouis:
Down to R15B3 since `cowboy` has no support that far back. @benoitc should I disable this test for these old platforms and skip it?

benoitc:
thanks! i am fine to disable the support of r16. i don't use it myself and next will probably be 17/18 only

jlouis:
I've disabled R15 support and this works! R16 is still in there. Next up, I'll clean up the commits by rebasing, then we can merge this.

rebar.config.script
```diff
@@ -21,7 +21,8 @@ Rebar2Deps = [{idna, ".*",
 
               {ssl_verify_hostname, ".*",
                {git, "https://github.com/deadtrickster/ssl_verify_hostname.erl",
-                {tag, "1.0.5"}}}],
+                {tag, "1.0.5"}}},
+              {cowboy, ".*", {git, "https://github.com/ninenines/cowboy.git", {tag, "1.0.4"}}}],
```
benoitc: maybe we should test the env here for example if a TEST has been set in the env. So we don't need to fetch cowboy all the time? Thoughts?

jlouis:
Rebased this to be a better and cleaner piece of work.

I agree we could test for the env if TEST has been set. It cut avoid the cowboy dependency by default. What is the idea in rebar 2.x? Is it already set, or do we have to supply it in some way when the script runs?

benoitc:
hrm I don't think anything exists in rebar2 . Imo we could juste set the
env and pass it by default. I can do the change anyway :)

jlouis:
Nice, then merge this. It is ready and partially solves #197 It doesn't do anything about the content disposition part of that issue.

edgurgel:
I wonder if we could use the existing `httpbin` instance that runs on travis for the integration tests (sadly they do not run on rebar3 https://github.com/benoitc/hackney/pull/202) instead of a new cowboy server.

benoitc:
@edgurgel good idea :)

jlouis:
You *can* run rebar3 if you put it into the repository itself and call it. The problem with this approach is the obvious: compiled a newer version of rebar3? You can't run it on the older releases and vice versa.

The other problem we face is that of including cowboy or not. The advantage of including it is you can run tests locally, since the `httpbin` instance might not be there locally... One has to make this work in order to make local development possible.

I've also seen integration tests are not running on `rebar3`. This is because there is no module named `hackney_integration` and hence the tests in `hackney_integration_tests` have no backing module and should not be run. Rebar 2.x does something "magical" here which frankly may have confused people more than it has helped.

benoitc:
So adding rebar3 inside travis can be done using such script:
https://github.com/barrel-db/hooks/blob/master/support/bootstrap_travis.sh

The script is launch before installing:
https://github.com/barrel-db/hooks/blob/master/.travis.yml#L5-L6

For HTTPBIN I guess it could also be automated but I see the point. Maybe we could replace httpbin then but it will require some work. Sometimes ago I wrote [echohttp:](https://github.com/benoitc/echohttp) which can be a good start.

I think it is a different issue anyway :) I propose to merge that PR, do the change in rebar config (checking if the environnement setting `TEST` is existing) and create a new issue to discuss further about it. Thoughts?

jlouis:
Yes, this change should go in, then get the ball rolling forward with fixes as they occur.

benoitc:
merged. Thanks!

PR #266, opened 2015-12-08
Changed typing of request to allow a url of the type url(), binary() and list()

sjaakieb:
We had some problems with the use of hackney in combination with dialyzer since the request methods had varying type signatures for the url.

### DIFF
--- a/src/hackney.erl
+++ b/src/hackney.erl
@@ -155,7 +155,7 @@ location(Ref) ->
 
 
 %% @doc make a request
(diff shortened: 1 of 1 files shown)


benoitc:
what is `url()` ?

sjaakieb:
From hackney.erl (line 43) :

-type url() :: #hackney_url{}.

benoitc:
hoh :+1:  sorry I forgot that one...

PR #297, opened 2016-03-14
Fix rebar.config.script for stable rebar3

tuscland:
That should hopefully handle all version number schemes used by rebar3 (sic).


benoitc:
merced with utf8 encoding issue fixed.

benoitc:
Thanks!

PR #298, opened 2016-03-14
Fix a typo in hackney_pool

tuscland:
Was calling metrics:update_histogram/3 without the first argument.


tuscland:
Sorry about the noise, I noticed someone else has proposed a similar change.

Issue #302, opened 2016-03-18
Error with Elixir and 1.5.4


zyro:
Upgrading from Hackney 1.5.2 to 1.5.4 crashes my Elixir 1.2.3 application with this error:

```elixir
** (Mix) Could not start application sonic: Sonic.start(:normal, []) returned an error: shutdown: failed to start child: :sonic_client_hackney_pool
    ** (EXIT) an exception was raised:
        ** (ArgumentError) argument error
            (stdlib) :ets.insert(:hackney_pool, {:sonic_client_hackney_pool, #PID<0.206.0>})
            (hackney) src/hackney_pool.erl:227: :hackney_pool.init/1
            (stdlib) gen_server.erl:328: :gen_server.init_it/6
            (stdlib) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
```

Perhaps an issue with starting the `:hackney_pool` ETS table?

benoitc:
did you start the application?

zyro:
Yes, it's on the list of applications being started.

benoitc:
can you share some code to helpe to reproduce the issue?

zyro:
Looks like the new `mix.exs` packaging doesn't define an application module to start. I'll put together a quick PR.

benoitc:
@zyro hackney 1.5.5 is released. Thanks for the patch!

zyro:
Awesome, thanks! :+1:

PR #303, opened 2016-03-18
Define application module in mix packaging

zyro:
Addresses #302.


benoitc:
thanks! will release ASAP

PR #325, opened 2016-06-17
Fix Elixir 1.4 warnings

milmazz:
Use parenthesis to remove the ambiguity with variables.

PR #326, opened 2016-06-18
Fix Elixir 1.4 warnings

whatyouhide:
For people using Elixir master (which is `1.4.0-dev` at the time of writing), `mix.exs` will warn because there are a couple of local 0-arity functions called without parentheses. This fixes that :)


edgurgel:
Apparently a duplicate of https://github.com/benoitc/hackney/pull/325

whatyouhide:
Whoops, completely missed that D: Sorry!

PR #336, opened 2016-07-13
support for UDS on Erlang 19, using 'http+unix' scheme

voltone:
I've only added limited test coverage for now. If I'm on the right track and this is something you'll consider merging I'll be happy to try and add some more tests and docs.

Note that I've had to override the Host header in requests I sent to the Docker engine, because it doesn't like the default Host header built by Hackney from the `netloc` of http+unix URLs. That could potentialy be handled in `host_header/2`.

### DIFF
--- a/src/hackney_local_transport.erl
+++ b/src/hackney_local_transport.erl
@@ -27,8 +27,8 @@ connect(Host, Port, Opts, Timeout) when is_list(Host), is_integer(Port),
 	(Timeout =:= infinity orelse is_integer(Timeout)) ->
 
     %% filter options
-    AcceptedOpts =  [linger, nodelay, keepalive, send_timeout,
-                     send_timeout_close, raw, inet6, reuseaddr,
+    AcceptedOpts =  [nodelay, keepalive, send_timeout,
+                     send_timeout_close, raw, reuseaddr,
                      ip, ip_address],
     BaseOpts = [binary, {active, false}, {packet, raw}],
     Opts1 = hackney_util:filter_options(Opts, AcceptedOpts, BaseOpts),
--- a/src/hackney_local_transport.erl
+++ b/src/hackney_local_transport.erl
@@ -1,0 +1,93 @@
+%%% -*- erlang -*-
+%%%
+%%% This file is part of hackney released under the Apache 2 license.
+%%% See the NOTICE for more information.
+%%%
+%%% Copyright (c) 2011-2012, Loïc Hoguin <essen@ninenines.eu>
+%%%
+-module(hackney_local_transport).
+-export([messages/1,
+         connect/3, connect/4,
+         recv/2, recv/3,
+         send/2,
+         setopts/2,
+         controlling_process/2,
+         peername/1,
+         close/1,
+         shutdown/2,
+         sockname/1]).
+
+%% @doc Atoms used to identify messages in {active, once | true} mode.
+messages(_) -> {tcp, tcp_closed, tcp_error}.
+
+connect(Host, Port, Opts) ->
+	connect(Host, Port, Opts, infinity).
+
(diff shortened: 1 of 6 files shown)


voltone:
I'll also have to review `AcceptedOpts` in hackney_local_transport.erl: the file is just a copy of hackney_tcp_transport.erl with line 37 changed to use `{local, Host}` as the destination, but I imagine not all `gen_tcp` options apply to local sockets.

benoitc:
@voltone thanks for the patch. Did you get further with the options? I tend to  think that most are usable but...

voltone:
@benoitc no, I haven't looked at it lately, I was waiting to hear whether you'd be interested in this feature. I'll have a look tonight, and try to add integration tests as well.

I've run into an issue with broken pipes, though: on my Mac, before the socket close notification I receive the text "Broken pipe", which obviously upsets the HTTP parser. Not sure if this is an issue of OTP 19's local socket beta or something that needs to be handled gracefully by Hackney...

voltone:
@benoitc: ok, so integration testing works on R19; do you have a suggestion for how I can suppress the local socket test case on <R19?

I'm not too worried about the "Broken pipe" issue for now: it seems to happen only when I put my laptop to sleep while a connection is open, and even then it is sporadic. Normal socket closure from client and server side work ok.

benoitc:
@voltone i would decorate the test using a macro. As for the build i'm not sure to disable it when < 19.  until the module build under old versions of Erlang I think it's fine.

src/hackney_local_transport.erl
```diff
@@ -0,0 +1,93 @@
...
+
+    %% filter options
+    AcceptedOpts =  [nodelay, keepalive, send_timeout,
+                     send_timeout_close, raw, reuseaddr,
+                     ip, ip_address],
+    BaseOpts = [binary, {active, false}, {packet, raw}],
+    Opts1 = hackney_util:filter_options(Opts, AcceptedOpts, BaseOpts),
+
+    %% connect
+    gen_tcp:connect({local, Host}, Port, Opts1, Timeout).
```
benoitc: shouldn't we pass a 0 port here?

benoitc: nvm it is.

benoitc:
@voltone 2c14417ce91191802cbd68f112657a1436ce5ec5 makes the test conditional. Sorry for the delay to merge the patch.

voltone:
@benoitc great, thanks! I just got back from vacation, hence the lack of responses from my side. About the port number, I guess we could ignore the value passed by the caller and force it to zero. It's just that the OTP docs leave a teeny tiny bit of ambiguity:

> In most API functions where you can use this address family the port number must be 0.

cybernetlab:
Hi, guys! What about absolute path names for socket files like `/var/run/sock`. When I try to do something like `hackney:request(get, <<"http+unix:///var/run/sock">>, [], <<>>, [])` the path `/var/run/sock` parsed as a URL path - not a URL host and I got error. If I compose `hackney_url` record manually all works fine and I use this method as a workaround now. But it would be nice to support above syntax. I think that some additional processing should be done in [hackney_url:parse_url](https://github.com/benoitc/hackney/blob/master/src/hackney_url.erl#L45) to catch such cases.

Should I create a new issue or it's enough to note it here?

Regards!

cjhowe:
Any update on this? Should I just go ahead and create an issue for the problem @cybernetlab brought up?

cybernetlab:
+1 - I'm still waiting for some comments

emilsoman:
@cybernetlab looks like the URL string needs to be URI encoded. "%2F" in the place of "/" should do it.

benoitc:
i'm not sure. sorry didn't notice that issue, since it was done on a closed PR... i will check if there is any issue. I think we should parse the scheme first before anything and then the parser adapted

PR #340, opened 2016-08-08
Update mix version to reflect the version increment to 1.6.1

nauger:
All the docs say that the current version is 1.6.1  Make this match what is on hex.pm.


benoitc:
thanks!

PR #353, opened 2016-10-20
prevent errors by handle trailing whitespace in header values

rich:
I was occasionally seeing errors in an Elixir project that I traced back to hackney. The most reliable way to see a failure was requesting http://www.apple.com Seems they sometimes are responding with a Content-Length value that had trailing whitespace.

This was the hackney error:
```
iex(1)> :hackney.get "http://www.apple.com"
** (ArgumentError) argument error
              :erlang.list_to_integer('27515     ')
    (hackney) /Users/rich/Projects/elixir/whistle/deps/hackney/src/hackney_response.erl:106: :hackney_response.wait_headers/4
    (hackney) /Users/rich/Projects/elixir/whistle/deps/hackney/src/hackney.erl:368: :hackney.send_request/2
```

I added a test directly into the `hackney_http` module because `parse_header/2` is not exported. If this is not the correct approach, I'm happy to fix; I'm new to both Elixir and Erlang.

### DIFF
--- a/src/hackney_http.erl
+++ b/src/hackney_http.erl
@@ -319,12 +319,8 @@ parse_header(Line, St) ->
     end,
     {header, {Key, Value}, St1}.
 
-parse_header_value(<< $\s, Rest/binary >>) ->
-    parse_header_value(Rest);
-parse_header_value(<< $\t, Rest/binary >>) ->
-    parse_header_value(Rest);
 parse_header_value(H) ->
-    H.
+    hackney_bstr:trim(H).
 
 
 parse_trailers(St, Acc) ->
@@ -525,3 +521,16 @@ get_property(content_type, #hparser{ctype=CType}) ->
     CType;
 get_property(location, #hparser{location=Location}) ->
     Location.
+
+%%% Private Tests
+-ifdef(TEST).
+
+    -include_lib("eunit/include/eunit.hrl").
+
+    parse_response_header_with_trailing_whitespace_test() ->
+        Response = <<"Content-Length: 27515  ">>,
+        Parser1 = parser([response]),
+        {header, {<<"Content-Length">>, Length}, _} = parse_header(Response, Parser1),
+        ?assertEqual(<<"27515">>, Length).
+
+-endif.


rich:
The CI checks that failed appear to be a failure on the CI side:

```
$ ./support/bootstrap_travis.sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  746k  100  746k    0     0  1013k      0 --:--:-- --:--:-- --:--:-- 1055k
===> Updating package registry...
===> Failed to write package index.
The command "./support/bootstrap_travis.sh" failed and exited with 1 during .
Your build has been stopped.
```

rich:
It does actually, that happened to be one of the cases I tested but, I guess it's not reflected in my test case:

```
iex(1)> :hackney_bstr.trim("\t  \tTesting  \t  ")
"Testing"
```

The Erlang `re` module docs seem to confirm http://erlang.org/doc/man/re.html "The \s characters are HT (9), LF (10), FF (12), CR (13), and space (32). "

benoitc:
hrm not sure why but in Erlang:

```erlang
1> hackney_bstr:trim(<<"\t  \tTesting  \t  ">>).
<<"Testing">>
2> hackney_bstr:trim(<<"\t essai \t essai">>).
<<"essai \t essai">>
```

rich:
Yeah, `trim/1` only handles leading or trailing whitespace. It's regex is defined like this `"^\\s+|\\s+$"`

Tried it in Erlang just to be certain we're speaking the same language:
```
Erlang/OTP 19 [erts-8.1] [source] [64-bit] [smp:4:4] [async-threads:0] [hipe] [kernel-poll:false] [dtrace]

Eshell V8.1  (abort with ^G)
1> hackney_bstr:trim(<<"testing\t123">>).
<<"testing\t123">>
2> hackney_bstr:trim(<<"\ttesting\t">>).
<<"testing">>
3> hackney_bstr:trim(<<"\ttesting\t123\t ">>).
<<"testing\t123">>
```

benoitc:
@rich yeah, i reread my example, and indeed it is working as expected. so merging the change, sorry for the noise, I sometimes need more café ....

PR #366, opened 2016-11-28
Handle continuation lines in HTTP headers correctly

obmarg:
[RFC2616 says](https://pretty-rfc.herokuapp.com/RFC2616#basic.rules):

> HTTP/1.1 header field values can be folded onto multiple lines if the
> continuation line begins with a space or horizontal tab. All linear
> white space, including folding, has the same semantics as SP. A
> recipient MAY replace any linear white space with a single SP before
> interpreting the field value or forwarding the message downstream.
>
> LWS            = [CRLF] 1*( SP | HT )
>
> SP             = <US-ASCII SP, space (32)>
> HT             = <US-ASCII HT, horizontal-tab (9)>

Hackney was not handling this correctly: it was stripping leading spaces
and tabs from continuation lines and then parsing them incorrectly as a
new header line.

This commit fixes that: `parse_header/1` now checks if the following
line starts with a space or tab and joins it onto the current line if
so.

### DIFF
--- a/src/hackney_http.erl
+++ b/src/hackney_http.erl
@@ -280,11 +280,11 @@ parse_header(#hparser{buffer=Buf}=St) ->
         [<<>>, Rest] ->
             {headers_complete, St#hparser{buffer=Rest,
                                           state=on_body}};
-        [<< " ", Line/binary >>, Rest] ->
-            NewBuf = iolist_to_binary([Line, Rest]),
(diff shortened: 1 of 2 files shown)


benoitc:
hrm the test  needs to have more thane one header to make sure it works :) But I will take of it. Thanks for the patch!

PR #370, opened 2016-12-06
correct hackney:connect example

JoshRagem:
Do not include the protocol in the `Host` or it will
be considered part of the host name--this results in an
`nxdomain` error.


benoitc:
thanks!

PR #389, opened 2017-02-28
fix duplicate content-type

sstrigler:
When sending requests (eg POST) with latest hackney as of 1.6.6 content-type headers got duplicated like:

Content-Type: application/vnd.myorg.profile-create-v1+json, application/vnd.myorg.profile-create-v1+json.

might very well be there need to be more fixes (multipart?)

### DIFF
--- a/src/hackney_request.erl
+++ b/src/hackney_request.erl
@@ -346,42 +346,44 @@ handle_body(Headers, ReqType0, Body0, Client) ->
                             {S, CT, Body0}
                         end,
   
+
+  Headers0 = hackney_headers:delete(<<"content-type">>, Headers),
   {NewHeaders, ReqType} = case {ReqType0, Body} of
                             {chunked, {file, _}} ->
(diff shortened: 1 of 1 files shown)


benoitc:
@sstrigler thanks for the patch! it definitely ggives a good glance on what need to be fixed. I think however we should go deeper and make the header handling more natural instead of doing back and forth, deleting/adding the same header. 

Last night I pushed the new Header module code in #390 to do this . I will plug the missing bit this morning, so expect a new release today (hopefully this morning).

benoitc:
let's keep the PR open however since it's pretty useful to figure important parts to fix.

sstrigler:
Perfect, thanks a lot!

benoitc:
@sstrigler fix landed in master. Thanks for your help anyway.

benoitc:
1.7.0 is available on 1.7.0

Issue #409, opened 2017-05-23
Hackney has dialyzer warnings


notriddle:
```
src/hackney.erl
 785: Function redirect/2 has no local return
 884: Record construction #hackney_url{scheme::'http' | 'http_unix' | 'https',raw_path::'undefined',path::bitstring(),qs::<<>>,fragment::<<>>,user::<<>>,password::<<>>} violates the declared type of field raw_path::binary()

src/hackney_http.erl
  92: Function parser/0 has no local return
 110: Function parser/1 has no local return
 111: Record construction #hparser{type::'auto',max_line_length::4096,max_empty_lines::10,empty_lines::0,state::'on_first_line',buffer::<<>>,version::'undefined',method::'undefined',partial_headers::[],clen::'undefined',te::'undefined',connection::'undefined',ctype::'undefined',location::'undefined',body_state::'waiting'} violates the declared type of field version::binary() and method::binary() and clen::integer() and te::binary() and connection::binary() and ctype::binary() and location::binary()
 173: The variable OK can never match since previous clauses completely covered the type {'error','bad_request'}
 176: The variable OK can never match since previous clauses completely covered the type {'error','bad_request'}
 210: The call hackney_http:parse_reason(#{}#, St::any(), Version::{integer(),integer()}, Acc::bitstring()) will never return since it differs in the 3rd argument from the success typing arguments: (any(), #hparser{type::atom(),max_line_length::integer(),max_empty_lines::integer(),empty_lines::integer(),state::atom(),buffer::binary(),version::binary(),method::binary(),partial_headers::any()],clen::integer(),te::binary(),connection::binary(),ctype::binary(),location::binary(),body_state::atom()}, binary(), binary())
 214: The call hackney_http:parse_reason(Rest::binary(), St::any(), Version::{integer(),integer()}, Acc::bitstring()) will never return since it differs in the 3rd argument from the success typing arguments: (any(), #hparser{type::atom(),max_line_length::integer(),max_empty_lines::integer(),empty_lines::integer(),state::atom(),buffer::binary(),version::binary(),method::binary(),partial_headers::any()],clen::integer(),te::binary(),connection::binary(),ctype::binary(),location::binary(),body_state::atom()}, binary(), binary())
 218: Function parse_reason/4 has no local return
[...]

PR #410, opened 2017-05-23
Describe, in the typespec, that URLs don't need every item defined

notriddle:
Part of #409


benoitc:
sorry i didn't see your pr until now. This has been fixed in latest master. Thanks a lot for the patch anyway !

PR #412, opened 2017-06-05
#411 sockets remain in CLOSE-WAIT state

### LINKED ISSUE #411: Too many sockets in CLOSE-WAIT state
I have a system which makes a large amount of async requests using hackney. Thank you for this library!
But I have a strange behaviour only for HTTP and not for HTTPS. When hackney works without default pool (proxy issues, btw), it does not close connection after request completion.

Script for this missbehaviour:

```
application:load(hackney).
application:set_env(hackney, use_default_pool, false).
application:ensure_all_started(hackney).

Url = "http://ipinfo.io/json".

Call = fun() -> {_, Ref} = hackney:get(Url, [], "", [{stream_to, self()}, async]), io:format("~p~n", [Ref]), timer:sleep(5000), flush(), hackney:close(Ref) end.

Call().
erlang:process_info(erlang:whereis(hackney_manager), links).

Call().
erlang:process_info(erlang:whereis(hackney_manager), links).

Call().
erlang:process_info(erlang:whereis(hackney_manager), links).
```

Each `Call()` adds one linked port to `hackney_manager`. These sockets are stuck in CLOSE-WAIT state and exhaust system limits. So I have to close them running external process each 10 minutes and calling `gen_tcp:close(Port)`

At the same time I don't see this behaviour for HTTPS (try with `Url = "https://ipinfo.io/json"`)

Zert:
Too rare case, only in async mode through proxy without connection pool.

Check this PR please to fix #411


benoitc:
I finally looked at the patch, thanks for it! But aren't you mixing different changes in the PR? 

The fix that cleanup the socket in `hackney_manager` should be enough to fix the issue, isn't it? The changes related to hackney_headers and header handling shouldn't be related. Can you open a PR for each changes? Also the changes in hackney_headers break the test. I'm wondering actually if it's really needed since the order should be already covered in `hackney_headers_new`.

Let me know.

Zert:
Ah, sorry, this is my changes. Not sure they should be in your repo at all.

PR #419, opened 2017-07-09
Add connect_timeout pool test

fishcakez:
Hackney pool leaks connections on connect timeout added in #402. The pool leaks the connections until the timed out process closes. Note that prior to that PR this test would also fail, or rather timeout, because it would block for an unbounded amount of time.

Issue #439, opened 2017-10-05
Response specifying connection is about to close is sometimes ignored


g-andrade:
Hi,

There appears to be an edge case in the way clean connection shutdowns are handled; sometimes, due to nondeterministic delays and concurrency, hackney is attempting to synchronously receive (`hackney_tcp:recv/2`) either the status line or the response headers and it gets `{error, closed}` because the connection has, well, closed.

Now, the issue here is that this may happen _even_ when we've gotten a valid HTTP response specifying that the connection was indeed about to close, but we're too late to get it. The end result is that the caller gets `{error, closed}` instead of the actual HTTP response.

The easiest way to reproduce this is to try executing a non-keepalive request with a *large* body, and possibly tweak the underlying connection's buffer to rather low values. 

I'm attaching a rebar3 project that reproduces the issue using hackney 1.9.0, cowboy and a pool of processes executing requests in parallel with payloads ranging on the 50 MiB range. To run it, one can execute `make run`, which will boot a rebar3 shell, and then execute ` hackney_error_closed:run().` In my machine I'm usually able to reproduce the issue after only a handful of requests.

[hackney_error_closed.zip](https://github.com/benoitc/hackney/files/1360184/hackney_error_closed.zip)

_It could be argued_ that cowboy / ranch might be closing the connections before the network stack has had a chance to flush pending data, but I've checked Wireshark and the full response _was_ being sent.

g-andrade:
Using async didn't help; most of the time the connection gets closed while I'm waiting either for the status line or the response headers. I suspect this is due to the way active mode is toggled back and forth in `hackney_stream`; it provides windows of opportunity for the connection to close without dispatching all the received data.

benoitc:
did you try with the `with_body` parameter?

g-andrade:
I did; the test case I attached to the opening comment replicates the issue using `with_body`. It's a tricky situation to solve - if one reacts to `{error, closed}` when calling `gen_tcp:send` / `ssl:send` before either A) setting the socket to active mode or B) doing so but nevertheless ignoring the process inbox, the response will be lost. I believe it's necessary to make the connection process aware of a 'half open' sort of state, wherein data can still be received but no longer sent - eventually the socket closure message will come in, but at least there's still a chance to receive the response.

benoitc:
@g-andrade sorry I missed the link. I will test. imo the right fix would be maintaining the socket in its own process . So it will continue to receive the sockets events and queue them locally (until some level).

[...]

PR #440, opened 2017-10-11
fix deadlock in hackney_pool during request timeout

jimdigriz:
This mostly resolves https://github.com/benoitc/hackney/issues/420 (and also includes https://github.com/benoitc/hackney/pull/419).

A race condition still exists, between the request timeout occurring till when the cast to `checkout_cancel` is processed.  I suspect it is really unlikely, though fortunately it is detectable so we could choose to do something about it later if this occurs.

A possible improvement would be to when the race occurs we dig around the `client` state element and reclaim the socket back into the pool?

### DIFF
--- a/src/hackney_pool.erl
+++ b/src/hackney_pool.erl
@@ -75,7 +75,10 @@ checkout(Host0, Port, Transport, #client{options=Opts}=Client) ->
         Transport}};
     {error, Reason} ->
       {error, Reason};
-    {'EXIT', {timeout, _}} -> {error, connect_timeout}
+    {'EXIT', {timeout, _}} ->
+      % socket will still checkout so to avoid deadlock we send in a cancellation
+      gen_server:cast(Pool, {checkout_cancel, {Host, Port, Transport}, RequestRef}),
+      {error, connect_timeout}
   end.
 
 %% @doc release a socket in the pool
@@ -304,6 +307,16 @@ handle_cast({set_maxconn, MaxConn}, State) ->
 handle_cast({set_timeout, NewTimeout}, State) ->
   {noreply, State#state{timeout=NewTimeout}};
 
+handle_cast({checkout_cancel, Dest, Ref}, State) ->
+  {Queues, Removed} = del_from_queue(Dest, Ref, State#state.queues),
+  case Removed of
+    true ->
+      NbWaiters = State#state.nb_waiters - 1,
+      {noreply, State#state{queues=Queues, nb_waiters=NbWaiters}};
+    false ->
+      % we leak the socket here but 'DOWN' will mop up for us when it times out
+      {noreply, dequeue(Dest, Ref, State)}
+  end;
 handle_cast(_Msg, State) ->
   {noreply, State}.
 
@@ -324,19 +337,7 @@ handle_info({ssl_error, Socket, _}, State) ->
 handle_info({'DOWN', Ref, request, _Pid, _Reason}, State) ->
   case dict:find(Ref, State#state.clients) of
     {ok, Dest} ->
-      Clients2 = dict:erase(Ref, State#state.clients),
-      case queue_out(Dest, State#state.queues) of
-        empty ->
-          {noreply, State#state{clients = Clients2}};
-        {ok, {From, Ref2}, Queues2} ->
-          NbWaiters = State#state.nb_waiters - 1,
-          _ = metrics:update_histogram(State#state.metrics,
-            [hackney_pool, State#state.name, queue_count], NbWaiters),
-          gen_server:reply(From, {error, no_socket, self()}),
-          State2 = State#state{queues = Queues2, clients = Clients2,
-            nb_waiters=NbWaiters},
-          {noreply, monitor_client(Dest, Ref2, State2)}
-      end;
+      {noreply, dequeue(Dest, Ref, State)};
     error ->
       {noreply, State}
   end;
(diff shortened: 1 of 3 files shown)


Hinidu:
@jimdigriz Did you see that it fails some test?

jimdigriz:
So does the master branch

Hinidu:
> So does the master branch

Hmmm, no, it fails not the test you've added but another one. Failed test is passed on master. See [Travis CI report](https://travis-ci.org/benoitc/hackney/builds/286401503?utm_source=github_status&utm_medium=notification).

src/hackney_pool.erl
```diff
@@ -304,6 +307,12 @@ handle_cast({set_maxconn, MaxConn}, State) ->
 handle_cast({set_timeout, NewTimeout}, State) ->
   {noreply, State#state{timeout=NewTimeout}};
 
+handle_cast({checkout_cancel, Dest, RequestRef}, State) ->
```
benoitc: How much should it be different from a 'DOWN' case ?
https://github.com/benoitc/hackney/blob/master/src/hackney_pool.erl

imo we should go back in the queue and trigger another waiting client in such case. Thoughts?

benoitc:
@jimdigriz mmm that tests seems related to the addition of the timeout error. Maybe a new edge case?

jimdigriz:
I'll try to have a nosey at this tonight.

benoitc:
@jimdigriz cool i will be available on irc tonight as well if you want to discuss some parts.

src/hackney_pool.erl
jimdigriz: Sounds like a good idea to me, reworking...

jimdigriz:
> >  So does the master branch

> Mmmm, no, it fails not the test you've added but another one. Failed test is passed on master. See Travis CI report.

The problem is with the unit test `hackney_pool_tests.erl`, it sets the pool connection size to 1 via the environment but never restores it...it should probably not be playing with the application environment anyway.

[Fixed the unit test](https://github.com/jimdigriz/hackney/commit/8032e2b0eaf3880fb436b0d09f23beced9690471).

jimdigriz:
> i like the idea. I think this should be handle like a 'DOWN' case though, so it triggers another waiting client.

I think this is causing the gen_server to do extra work when we know `queue_out()` will always return `empty`; as we are cancelling a queued `checkout`.

For example we have `MaxConn` set to 10, we try to queue 11, it times out and we send in a cancellation.  Meanwhile `length(clients)=10, length(queue)=1`.  We sending in `checkout_cancel` and we will still have `length(clients)=10` as we are cancelling the request that is in `queue`.

Does this sound right?

benoitc:
@jimdigriz mmm right you need to clean yourself, but I think we still need a way to trigger another waiting connection. Ie since you reduce the number of connections to be be released, let's handle one another from the queue and timeout ourself. Does it make sense?

jimdigriz:
We are only removing requests from the `queue` dictionary, all of those are just awaiting for an available socket.  If a trigger could do something, it would have already from another socket, right?

What we can do though is fix the race by recovering the socket if we see that the request has moved from the `queue` into the `clients` dictionary.  Here we can see if someone else may be interested in the socket.

jimdigriz:
What actually calls `DOWN`?  I am unable to find any `monitor()`/`link()`/`process_flag(trap_exit, true)` calls in `hackney_pool.erl`.

benoitc:
@jimdigriz this is the manager: https://github.com/benoitc/hackney/blob/master/src/hackney_manager.erl#L408

This part is inherited from the past. The manager is responsible of maintaining the requests. This is the part i expect to replace soon by maintainting each connection states in their own process instead (ie generalising the async case). Which would also solve #439 .

I was thinking this cancel was doing exactly what you do right now somehow but preemptively . The reason to trigger (queuing out) after having deleted yourself is to enable another waiter to receive a socket. If not it will wait until another socket is released. Isn't it?

jimdigriz:
Okay, figured out `no_socket` actually triggers the creation of a socket, no longer panic'ing :)

src/hackney_pool.erl
```diff
@@ -304,6 +307,16 @@ handle_cast({set_maxconn, MaxConn}, State) ->
 handle_cast({set_timeout, NewTimeout}, State) ->
   {noreply, State#state{timeout=NewTimeout}};
 
+handle_cast({checkout_cancel, Dest, Ref}, State) ->
+  {Queues, Removed} = del_from_queue(Dest, Ref, State#state.queues),
+  case Removed of
+    true ->
+      NbWaiters = State#state.nb_waiters - 1,
+      {noreply, State#state{queues=Queues, nb_waiters=NbWaiters}};
```
benoitc: why not dequeuing it there too?

jimdigriz: As we have non-zero `nb_waiters` only want `length(clients)==MaxConns` right, otherwise when one of those clients finished, `hackney_pool` would have dequeued already?

benoitc: releasing a waiter only happen when a connection is put back in the loop or when a request is canceled: forcefully closed via  hackney:close/1 or when a process actually requesting crash. I think we are in the second case there, so imo we should dequeue one of the waiter to actually tell him to create a new request imo.

jimdigriz: We will exceed `pool_size` though?

benoitc: @jimdigriz right. applying your change now. Will release sometimes today :)

PR #442, opened 2017-10-11
Fix memory leak in hackney_manager.

### LINKED ISSUE #441: Memory leak in hackney_manager when using async responses.
I've been doing some load testing of an Elixir service we've written, that makes
a lot of HTTP requests with hackney, using async responses. I noticed that after
about a million requests, the hackney_manager process ended up using about 500MB
of memory. This memory use did not go down when I turned the load testing off,
and the application was sitting totally idle.

Looking into the state of the process, it's pid dict contained about a million
entries - presumably one for every request that had been made.

It's quite easy to reproduce this using HTTPoison in elixir:

```
iex(1)>   HTTPoison.get("http://www.google.com", [], stream_to: self)
iex(2)> {_, pids, _} = :sys.get_state(:hackney_manager)
iex(3)> :dict.size pids
1
iex(4)>   HTTPoison.get("http://www.google.com", [], stream_to: self)
iex(5)> {_, pids, _} = :sys.get_state(:hackney_manager)
iex(6)> :dict.size pids
2
```

obmarg:
I've been doing some load testing of an Elixir service we've written, that makes
a lot of HTTP requests with hackney using async responses. I noticed that after
about a million requests, the hackney_manager process ended up using about 500MB
of memory. This memory use did not go down when I turned the load testing off,
and the application was sitting totally idle.

Looking into the state of the process, it's pid dict contained about a million
entries - presumably one for every request that had been made.

After investigating, I've tracked this down to the handle_exit implementation
for stream processes.  It was not actually trying to remove the owner PID from
the dict, but was trying to remove the stream PID twice.

Fixes #441

### DIFF
--- a/src/hackney_manager.erl
+++ b/src/hackney_manager.erl
@@ -503,7 +503,7 @@ handle_exit(Pid, {Ref, stream}, Reason, State) ->
     [{Ref, {Owner, Pid, #request_info{pool=Pool}=Info}}] ->
       %% unlink the owner
       unlink(Owner),
-      Pids2 = dict:erase(Pid, Pids1),
+      Pids2 = dict:erase(Owner, Pids1),
       %% if anormal reason let the owner knows
       _ = case Reason of
             normal ->  ok;


benoitc:
looks good oto me. Do you have by chanc a test to reproduce it?

src/hackney_manager.erl
```diff
@@ -503,7 +503,7 @@ handle_exit(Pid, {Ref, stream}, Reason, State) ->
     [{Ref, {Owner, Pid, #request_info{pool=Pool}=Info}}] ->
       %% unlink the owner
       unlink(Owner),
-      Pids2 = dict:erase(Pid, Pids1),
+      Pids2 = dict:erase(Owner, Pids1),
```
benoitc: actually i'm wondering what happen when a owner listen on 2 streams. Maybe it shouldn't be removed? I will test it.

obmarg:
There's reproduction steps in the linked issue, but I don't have a unit test right now.  All my testing was done by hand.

I can try and write a unit test in the next few days though, are there any existing hackney_manager tests I can use to help?

benoitc:
@obmarg writing a simple test based on what you provided that should be enough :) there is no real testing yet.. Actually using your steps and adding an additional stream on the same process should allow us to test what I commented.

src/hackney_manager.erl
obmarg: Hmm, that is a good point.  Though if an owner can listen on 2 streams does that not make the whole idea of having a dict of owner-pid -> stream data a bit flawed?

obmarg:
@benoitc Cool - I'm extremely busy with some other stuff right now, but I'll try and get round to that soon.

src/hackney_manager.erl
benoitc: @obmarg I think that's why you have the owner and the stream. This is something I want to fix anyway. We should probably only manage streams and link them directly to their owners...

benoitc:
following our discussion I pushed f679aca15439e6e631d5e9d54baf010848d85915 that fixes the owner tracking. I also added the test. Thanks for the PR and the chat anyway!

obmarg:
Thanks @benoitc - the fix in f679aca does seem to have fixed the issue I was experiencing.  However, I'm now seeing some weird crashes in hackney_manager that take down a bunch of my processes.

I've not done _much_ investigation just yet though, so this could definitely be my bad.  I'll raise an issue if it ends up looking like a hackney issue.

benoitc:
@obmarg if you havec some time to share the traces/logs let me know. You can send them privately if you want.

obmarg:
@benoitc I'm seeing this from the `hackney_manager`, after a process that was running an async request raises an exception:

```
12:31:00.140 [error] GenServer :hackney_manager terminating
** (MatchError) no match of right hand side value: {:error, :normal}
    (hackney) /Users/graeme/src/securi/deps/hackney/src/hackney_manager.erl:524: :hackney_manager.clean_requests/5
    (hackney) /Users/graeme/src/securi/deps/hackney/src/hackney_manager.erl:498: :hackney_manager.handle_owner_exit/4
    (stdlib) gen_server.erl:616: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:686: :gen_server.handle_msg/6
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
```

Then this from a ton of other processes that are running requests:

```
12:31:00.141 [error] Ranch protocol #PID<0.530.0> (:cowboy_protocol) of listener Securi.Router.HTTP terminated
** (exit) an exception was raised:
    ** (MatchError) no match of right hand side value: {:error, :normal}
        (hackney) /Users/graeme/src/securi/deps/hackney/src/hackney_manager.erl:524: :hackney_manager.clean_requests/5
        (hackney) /Users/graeme/src/securi/deps/hackney/src/hackney_manager.erl:498: :hackney_manager.handle_owner_exit/4
        (stdlib) gen_server.erl:616: :gen_server.try_dispatch/4
        (stdlib) gen_server.erl:686: :gen_server.handle_msg/6
        (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
```

I've no easy steps to reproduce just yet though: so far I'm only seeing this when running a high volume of requests through my service.

obmarg:
`hackney_sup` seems to be doing something weird now as well:

image

This is from a system that should have been idle for > 24 hours, but you can see there's a ton of messages in the `hackney_sup` queue, and the # of reductions is continually going up.

benoitc:
@obmarg 1ab27d4b40452acb7902528a30c8eccbcff2c7f9 should fix the issue. Let me know.

PR #443, opened 2017-10-18
fix owner tracking

### LINKED ISSUE #441: Memory leak in hackney_manager when using async responses.

benoitc:
previously we were unlinking a process that started a request each time
a stream were closing or when we were streaming to another process the
response.

This change allows a process to start multiple requests and stream. It
fix it by doing the following:

* only streams and owners are supervised by the manager to track the
socket life
* the stream is responsible of monitoring its target. If the target
process die it exists and the manager clean it
* a process can now have multiple client referenced to it. (which is
done via the addition of track_owner/3 and untrack_owner/3 functions in
the manager).

fix #441

### DIFF
--- a/src/hackney_manager.erl
+++ b/src/hackney_manager.erl
@@ -410,7 +410,7 @@ handle_info({'EXIT', Pid, Reason}, State) ->
       handle_stream_exit(Pid, Ref, Reason, State);
     {ok, Refs} when is_list(Refs) ->
       handle_owner_exit(Pid, Refs, Reason, State);
-    _ ->
+    _Else ->
       {noreply, State}
   end;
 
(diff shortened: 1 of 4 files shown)


benoitc:
should replace #442 @obmarg can you test?

PR #444, opened 2017-10-18
Ignore empty port value on redirect

tnt-dev:
Some servers return redirect codes with Location header like "http://example.com:/
In this situation we should just use standard 80/443 port.


benoitc:
thanks!

Issue #445, opened 2017-10-19
Hackney leaks manager refs 


scohen:
I updated to master two days ago and observed a tremendous number of client references in the `hackney_manager_refs` ets table.
I've been able to reproduce quite easily, to the point where I wonder if there's something I'm doing incorrectly, but the documentation hasn't changed. 

The following test shows the problem
```elixir
def t do 
  before = :ets.info(:hackney_manager_refs, :size)
  {:ok, _, _, client} = :hackney.get("http://www.google.com")
  during = :ets.info(:hackney_manager_refs, :size)
  {:ok, _body} = :hackney.body(client)
  after = :ets.info(:hackney_manager_refs, :size)
  {before, during, after}
end

> HackneyTest.t
{0, 1, 1}  # Under 1.10.0
```

Under 1.9.0 we get `{0, 1, 0}`, which is the correct response. 
I used git bisect and f679aca15439e6e631d5e9d54baf010848d85915 is the commit that introduced the problem.

benoitc:
should be fixed in latest master. Thanks for the ticket!

PR #446, opened 2017-10-19
Added unit tests for reference leaks

scohen:
I added two tests, one of which would have caught the manager leak in #445.


benoitc:
commited. thanks!

PR #450, opened 2017-10-24
Handle empty or invalid ‘Content-Length’

tnt-dev:
Ignore non-integer ‘Content-Length’ header.

### DIFF
--- a/src/hackney_http.erl
+++ b/src/hackney_http.erl
@@ -293,8 +293,10 @@ parse_header(Line, St) ->
                  end,
   St1 = case hackney_bstr:to_lower(hackney_bstr:trim(Key)) of
           <<"content-length">> ->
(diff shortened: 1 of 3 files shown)


src/hackney_http.erl
```diff
@@ -293,8 +293,10 @@ parse_header(Line, St) ->
                  end,
   St1 = case hackney_bstr:to_lower(hackney_bstr:trim(Key)) of
           <<"content-length">> ->
-            CLen = list_to_integer(binary_to_list(hackney_bstr:trim(Value))),
-            St#hparser{clen=CLen};
+            case catch list_to_integer(binary_to_list(Value)) of
```
benoitc: any reason to not use `binary_to_integer/1` ? though not sure which version of erlang it appeared ...

benoitc:
Thanks for the patch.  I will put a note for later to go to binary_to_integer in next major version :)

Issue #452, opened 2017-10-27
case_clause issue on cancel_checkout


byronpc:
I'm receiving this crash on my server

`{case_clause,0},[{hackney_pool,handle_cast,2,[{file,"src/hackney_pool.erl"},{line,312}`

Not sure if this is related to a memory leak we're experiencing when this crash occurs.

Upon checking the code, I can see `del_from_queue` can return `{Queues, 0}` where 0 is supposed to be a boolean for `Removed`. I think it should be `false` right?

byronpc:
@jimdigriz

jimdigriz:
Eugh, badly refactored, sorry!

When I moved from passing back an integer to a boolean I forgot to fix up the `error` path.  Can you try https://github.com/benoitc/hackney/pull/453

byronpc:
@jimdigriz there seems to be an issue with checkout_cancel in general. I'm not sure where but it seems to get stuck in the messagequeue of the pool from time to time causing all succeeding checkouts to the pool to fail.

jimdigriz:
...stuck for ~8 seconds (the default for `connect_timeout` unless you have it set to something else) before being immediately replaced with new `checkout_cancel` cast message?

byronpc:
the thing is, i'm not sure if the message queue stacking up happens because of load. If it is because of load, I will occasionally see the message queue warnings and timeouts growing. But in this case it happens only once in a while (like once in an hour or so) but when it happens, all succeeding calls will timeout (like i get thousands of timeouts and only lasts for around 2 seconds) and the message queue error will be full of checkout_cancel messages and seems lots of calls to `del_from_queue`, then it will recover. Again, I'm not sure what happens internally but this is my observation. The issue may be from another place that holds up the gen_server and the `checkout_cancel` messages are only a side effect. Will look more into this.

jimdigriz:
@byronpc how do you expect hackney to behave if you have a connect timeout set to (for example) 8 seconds, the pool size limited to 10 and the response timeout set to 60 seconds?  What should hackney do if the upstream HTTP server, given these constraints you may have set, takes 30 seconds to reply whilst you are sending ~100 requests per second?

The problem is the *upstream* HTTP server is under load and struggling, not hackney.

jimdigriz:
@byronpc did the https://github.com/benoitc/hackney/pull/453 fix the `case_clause` for you?

PR #453, opened 2017-10-27
fix return value of del_from_queue

jimdigriz:
Fixes https://github.com/benoitc/hackney/issues/452 ....probably.

### DIFF
--- a/src/hackney_pool.erl
+++ b/src/hackney_pool.erl
@@ -475,7 +475,7 @@ add_to_queue({_Host, _Port, _Transport} = Dest, From, Ref, Queues) ->
 del_from_queue({_Host, _Port, _Transport} = Dest, Ref, Queues) ->
   case dict:find(Dest, Queues) of
     error ->
-      {Queues, 0};
+      {Queues, false};
     {ok, Q} ->
       Q2 = queue:filter(fun({_, R}) -> R =/= Ref end, Q),
       Removed = queue:len(Q) =/= queue:len(Q2),


byronpc:
@jimdigriz I've been observing, the checkout_cancel in general. And it seems this cast gets stuck and causes a deadlock in the pool. 

```
Message queue:[{'$gen_cast',{checkout_cancel,{"gcm-http.googleapis.com",443,hackney_ssl},#Ref<0.0.7340050.2138>}},{'$gen_
cast',{checkout_cancel,{"gcm-http.googleapis.com",443,hackney_ssl},#Ref<0.0.7340050.2114>}},{'$gen_cast',{checkout_cancel
,{"gcm-http.googleapis.com",443,hackney_ssl},#Ref<0.0.7340050.2147>}},{'$gen_cast',{checkout_cancel,{"gcm-http.googleapis
.com",443,hackney_ssl},#Ref<0.0.7340050.2149>}},{'$gen_cast',{checkout_cancel,{"gcm-http.googleapis.com",443,hackney_ssl}
,#Ref<0.0.7340050.2151>}},{'$gen_cast',{checkout_cancel,{"gcm-http.googleapis.com",443,hackney_ssl},#Ref<0.0.7340050.2156
>}},{'$gen_cast',{checkout_cancel,{"gcm-http.googleapis.com",443,hackney_ssl},#Ref<0.0.7340050.2127>}},{'$gen_cast',{chec
kout_cancel,{"gcm-http.googleapis.com",443,hackney_ssl},#Ref<0.0.7340050.2136>}},{'$gen_cast',{checkout_cancel,{"gcm-http
.googleapis.com",443,hackney_ssl},#Ref<0.0.11272197.128945>}},{'$gen_call',{<0.25658.2592>,#Ref<0.0.18350081.17329>},{che
ckout,{"gcm-http.googleapis.com",443,hackney_ssl},<0.25658.2592>,#Ref<0.0.18350081.17315>}},{'$gen_call',{<0.10419.2592>,
#Ref<0.0.18350081.17330>},{checkout,{"gcm-http.googleapis.com",443,hackney_ssl},<0.10419.2592>,#Ref<0.0.18350081.17317>}}
,{'$gen_call',{<0.3498.2409>,#Ref<0.0.18350081.17338>},
```

As you notice there's a couple of checkout_cancel messages in the queue causing all other succeeding calls to hackney to fail for a short time.

jimdigriz:
You say 'fail for a short time', so I am going on the assumption that that message is not being handled is `hackney_pool` is busy elsewhere; its a gen_server, and nothing exciting is going on here.

This behavior is expected if the upstream HTTP server is taking its time to respond (longer than the set `connect_timeout`) which is the way the state machine gets to this point.

The patch does not fix this, and it is impossible to fix this if you want to enforce a limit on `pool_size` as the problem is the upstream server timing out or at least being very slow.  What the `checkout_cancel` patch fixed was that hackney would deadlock and not recover from this situation.

If your environment is where timeouts/slow-responses are the norm *and* you want to limit the `pool_size` there is nothing that you can do here...  Maybe you can lessen the effect by setting `connect_timeout` to ~3x the RTT to the upstream server you are contacting?

silviucpp:
@jimdigriz we are experience this issue 100 % when we increase the pool size. For example with a pool size of 100 connections we didn't experienced the problem for more than 1 day, with 200 connections we experienced the problem after 8 hours and with 500 connections it's going crazy in less than 2 hours (can be 100 % replicated)

So somehow the pool size have an important role in this problem

Silviu

jimdigriz:
....because your upstream HTTP server you are querying, which regularly struggles with <100 connections, is now completely swamped with 5x the load.  The patch fixes it so that once the upstream HTTP server recovers, hackney *also* recovers.

I am making a really poor job of communicating here.  All I can recommend is you plot a histogram (via a packet capture and not from inside erlang) showing the HTTP response times of your HTTP server.  Maybe you have access to the upstream server and can make the measurements from there...if you can also look at the TCP listen queue drops for hints.

silviucpp:
Hello @jimdigriz,

The server we are query is Google GCM service (gcm-http.googleapis.com). Idea is that we started to face this problems after we upgraded hackney from 1.4.x to last stable. Our traffic didn't increased in number of pushes delivered for this reason we suspect something with hackney

Anyway probably we need to enable hackney stats and to see the response times 

Thanks,
Silviu

jimdigriz:
As it is Google, rather than overloading, it could be throttling which would have the same effect.

sntran:
I cannot produce a standalone repo, but I do have the same issue when working with Google Maps API. The library is at https://github.com/sntran/ex_maps, when running tests, work fine with `hackney@1.9`, but generates the exception in 1 out of 40+ tests when using `hackney@1.10.1`

Last message: `{:"$gen_cast", {:checkout_cancel, {'maps.googleapis.com', 443, :hackney_ssl}, #Reference<0.1841845360.644349953.119089>}}`

PR #454, opened 2017-10-30
Fix error with incorrect domain handling in SOCKS5 response

Zert:
Fix bad_request error when SOCKS5 proxy replied with domain name


benoitc:
thx!  i should find a way to add a test for it also :)

lattenwald:
This thing breaks socks5 proxy support. I created issue #460

Issue #449, opened 2017-10-20
Hackney doesn't send SNI when proxying


dspencer001:
Hackney: 1.8.0 (it appears on 1.10.1 as well)
Erlang: 19.3
Elixir: 1.3.4

When doing a GET to https://nationalinterest.org/ through squid as a proxy, Hackney fails with a tls_alert error because the client handshake failed. It looks like it doesn't send the SNI when proxying.

I couldn't find the ssl option to send the SNI anywhere in the code, so I have no idea how it works when not proxying.

Adding `{server_name_indication, Host}` to the options returned by ssl_opts in hackney_connect fixed it locally.

I can open a PR adding the option to ssl_opts if the SNI can always be sent.

To reproduce:
Run squid (defaults to port 3128).

`:hackney.get("https://nationalinterest.org/", [], <<>>, [proxy: "localhost:3128"])`

Hackney responds with `{:error, {:tls_alert, 'handshake failure'}}`

benoitc:
@dspencer001 thank for the report. Only issue with this option is that it's only available on erlang > 19 so we need a way to handle it. I'm thinking to add a flag via rebar.config for now. Thoughts?

dspencer001:
I'm not very familiar with erlang, so I don't know what would be best. It looks like the option is supported on erlang > 17 though according to the [docs.](http://erlang.org/documentation/doc-6.0/lib/ssl-5.3.4/doc/html/ssl.html)

dspencer001:
I would be happy to open a PR though if you can point me to an example of how you'd like it done.

PR #455, opened 2017-11-07
Send SNI for erlang >= 17

dspencer001:
I was unable to test this on R16B03-1. It should work though. Fixes https://github.com/benoitc/hackney/issues/449


benoitc:
thanks!

Issue #155, opened 2015-01-19
Add transparent Content-Encoding: gzip/deflate support


seriyps:
So, should add `Accept-Encoding: gzip,deflate` to request, then check for `Content-Encoding: gzip` or `Content-Encoding: deflate` and decompress body (after de-chunking).
Decompression may be done in single pass (`zlib:gunzip/1` for gzip and `zlib:unzip/1` for deflate) in case of synchronous requests and using streaming (`zlib:open/0, zlib:inflateInit/1, zlib:inflate/2 ..., zlib:close/1`) in case of asynchronous.
Single-pass example can be found here https://github.com/seriyps/xhttpc/blob/master/src/middlewares/compression_middleware.erl

Some option should be added to enable this feature.

benoitc:
will add it in next release. Thanks for the suggestion.

zweizeichen:
+1

seriyps:
Also, bear in mind new Erlang 18.0 API to protect from gzip bombs http://www.erlang.org/documentation/doc-7.0-rc1/erts-7.0/doc/html/zlib.html#inflateChunk-2

benoitc:
feature is coming. I am in the middle of adding some big improvments to hackney. I will probably include it right after.  Thanks for the hint.

tuscland:
+1 :)

zyro:
+1 Would be great to see this handled transparently for both requests and responses.

huoshiqiu:
+1 :)

Anonyfox:
+1 on this :)

kostya:
+1

edgurgel:
Can I help with this? I would be glad to follow some guidance and help with this feature.

benoitc:
@edgurgel thanks! It should probably be added to the function receiving the body, if you wrap them and keep the state it should be OK. Anyway betetr to wait over the we, with the new recv and recv_multipart functions :)

tlvenn:
Any update on this feature ?

loicvigneron:
+1 :)

stevegraham:
What's the status of this issue?

benoitc:
No work have been done on my side yet. But any help is appreciated. Anyway I expect that this feature will land after the current rework of the internals.

sunboshan:
[...]

PR #456, opened 2017-11-16
Transparent (content-encoding) compression

jimdigriz:
An initial stab at merging my wrapper into hackney its-self

https://github.com/benoitc/hackney/issues/155


src/hackney_http.erl
```diff
@@ -325,7 +328,40 @@ parse_trailers(St, Acc) ->
     _ -> error
   end.
 
-parse_body(St=#hparser{body_state=waiting, te=TE, clen=Length,
+-define(MAX_WBITS, 15).	% zconf.h
+parse_body(St=#hparser{body_state=waiting, ce=CE, clen=Length, method=Method}) ->
+  St2 = case CE of
```
seriyps: I think it would be nice to use `parse_body` function clauses instead of this case

src/hackney_pool.erl
```diff
@@ -475,7 +475,7 @@ add_to_queue({_Host, _Port, _Transport} = Dest, From, Ref, Queues) ->
 del_from_queue({_Host, _Port, _Transport} = Dest, Ref, Queues) ->
   case dict:find(Dest, Queues) of
     error ->
-      {Queues, 0};
+      {Queues, false};
```
seriyps: What is it about?

src/hackney_http.erl
```diff
@@ -325,7 +328,40 @@ parse_trailers(St, Acc) ->
...
+  parse_body2(St2);
+parse_body(St=#hparser{encoding={zlib,Z}}) ->
+  case parse_body2(St) of
+    {ok, Chunk, St2} ->
+      Chunk2 = iolist_to_binary(zlib:inflate(Z, Chunk)),
+      {ok, Chunk2, St2};
+    {done, Rest} ->
+      Rest2 = iolist_to_binary(zlib:inflate(Z, Rest)),
+      ok = zlib:inflateEnd(Z),
+      ok = zlib:close(Z),
```
seriyps: Are you sure this is the only condition when zlib handle might be closed? Is it possible to leave some zlib handles stale in case of errors or smth like that?

seriyps:
Tests a failing because of a dialyzer warning:

```
_build/default/lib/hackney/src/hackney_http.erl
 111: Record construction #hparser{type::'auto',max_line_length::4096,max_empty_lines::10,empty_lines::0,state::'on_first_line',buffer::<<>>,method::<<>>,partial_headers::[],te::<<>>,ce::<<>>,connection::<<>>,ctype::<<>>,location::<<>>,body_state::'waiting',encoding::'undefined'} violates the declared type of field encoding::{atom(),_}
 523: Function parse_options/2 will never be called
```

I would suggest you to change type of `encoding` field to `{atom(), any()} | undefined`.

src/hackney_pool.erl
jimdigriz: https://github.com/benoitc/hackney/pull/453, if there is the chance this PR gets close to being merged, it can be dropped (or rather merging https://github.com/benoitc/hackney/pull/453 is needed anyway so this solves its-self)

src/hackney_http.erl
jimdigriz: Maybe, but I was following style hints from @benoitc in the handling of `te` which is handled with a `case`.  Personally I am not bothered.

src/hackney_http.erl
jimdigriz: Nope, pretty sure we leak if something implodes (ie. server prematurely closes stream).

Really at this point more interested in:
 * does this work for others
 * anyone see any regressions
 * does `compress` as an option suit the users of hackney
 * is there a better layer for me to plumb this in or less intrusively

Fixing bugs/edge cases I suspect would not change drastically the shape of this patch but changing the plumbing would, as would maybe some fixes for regressions.  Make sense?

I dislike wasting time if I am on the wrong footpath and no one tells me :)

fenollp:
Shouldn't the new `compress` option be inferred from the `accept` header instead of needing to be explicitly specified? Was it not done this way for backwards compatibility reasons?

jimdigriz:
I used `compress` to avoid having to create a API call so then the application can tell *what* to put in its headers (today we support `deflate`/`gzip` simply as it brings in no dependencies).  I suspect to a developer most of the time they think more "compress where possible" rather than "please only use deflate".

However, this is how *I* think of using `hackney`, if I need something else, I just hack the source and submit a PR, others I am willing to accept might prefer something else :)

fenollp:
Works great for me. Thanks :)

src/hackney_http.erl
```diff
@@ -325,7 +328,40 @@ parse_trailers(St, Acc) ->
     _ -> error
   end.
 
-parse_body(St=#hparser{body_state=waiting, te=TE, clen=Length,
+-define(MAX_WBITS, 15).	% zconf.h
+parse_body(St=#hparser{body_state=waiting, ce=CE, clen=Length, method=Method}) ->
```
seriyps: I think we have some inconsistency here: body will be unzipped regardless of `compress` option. But maybe we should check that both `Content-Encoding` header and `compress` option are active to decompress the body?
Because, with your current implementation, there will be no way at all to get compressed body as is - without decompression. For example if I need to get compressed body for some reason (eg, send it further or store on disk without looking into it... proxy...) I may set `Accept-Encoding` manualy and will get compressed body in response. But if I need transparent compression, I'll use `compress` option.

jimdigriz: Good catch, I'll get back with a fix in a day or two.

benoitc: @jimdigriz did you fix it?

jimdigriz: I took a stab at it but noticed that to push the 'do transparent compression' flag into `#hparser{}` is quite invasive.  I do plan to return to this but real life has caught up with me. :)

benoitc: hrm which also let me think i should revisit that data structure, it's quite duplicating the `#client{}` one ...

cgorshing: I really appreciate the work you are doing, this is something that I am needing. A couple questions for you.
So the `compress` option  means both to compress the request _and_ to decompress the response? I see these as two different concerns. Maybe I've missed something reading through the comments.
I agree with seriyps above, I would only want to decompress the response if I pass in the option.
One one case that I have, I don't really care about my request being compressed, but I care about the response being uncompressed. When I see the option `compress` I think there might be another option for the response. Thinking out loud.
Appreciate the work you have done, I'm wanting to get this into prod soon myself.

jimdigriz: I agree completely too with @seriyps thoughts too, it was something I had not thought of at the time.  As @benoitc noticed too, to resolve this will probably involve a substantial change and I have not really looked much into it since.

@benoitc my concern is only wasting everyones time.  If you can describe the changes you would like to see to replumb this, when I find the time to progress this, I would then be able to move in a direction that works for you.

src/hackney_http.erl
benoitc: "does compress as an option suit the users of hackney. " yes

src/hackney_http.erl
fenollp: @benoitc Any idea how to help @jimdigriz ^?

src/hackney_http.erl
jimdigriz: updated

src/hackney_http.erl
seriyps: @jimdigriz I believe this was resolved with the last commit? Unfortunately, I can't see what changed in the last commit because of force-push..

jimdigriz:
Finally found the time to handle the `compress` flag properly; I moved the functionality from `hackney_http` into `hackney_response` which seems more appropriate as Content-Encoding is application layer feature rather than part of the transport.

I went to try adding unit tests, but looking around it seems to be non-trivial to add gzip/deflate support to gunicorn so got stopped in my tracks. Any suggestions on what to do here?

jadeallenx:
@benoitc Any update on this?

PR #457, opened 2017-11-20
Hackney performance improvements

scohen:
Hackney has been responsible for taking down our site a couple of times, this fixes some of the issues:

1. Hackney's pool runs at high priority despite the fact that docs explicitly tell users not to use this flag. This causes issues when Erlang's scheduler spends too much time interacting with the pool and not enough time doing other things. We've been running without this flag for a couple months, and there's no difference in Hackney's performance. 
1. There was a case clause mismatch that was spamming our logs.
1. Hackney utilizes a metrics module that abstracts between several different metrics providers. This module is dynamically recompiled every time a new connection is created. This recompilation is incredibly expensive, requiring every process in the system to be woken up from hibernation and then send a message. The code server then needs to receive every message in order to continue. For large pools, this can take *minutes* to complete and during this time, Erlang becomes totally unresponsive. The fix is to move compilation to application start and to read an application variable to find the metrics module.

### DIFF
--- a/src/hackney_app.erl
+++ b/src/hackney_app.erl
@@ -20,6 +20,8 @@
 %% ===================================================================
 
 start(_StartType, _StartArgs) ->
+  Metrics = metrics:init(hackney_util:mod_metrics()),
+  application:set_env(hackney, metrics, Metrics),
   hackney_sup:start_link().
 
 stop(_State) ->
--- a/src/hackney_connect.erl
+++ b/src/hackney_connect.erl
@@ -66,7 +66,7 @@ create_connection(Transport, Host, Port, Options, Dynamic)
   MaxBody = proplists:get_value(max_body, Options),
 
   %% get mod metrics
-  Engine = metrics:init(hackney_util:mod_metrics()),
+  {ok, Engine} = application:get_env(hackney, metrics),
 
   %% initial state
   InitialState = #client{mod_metrics=Engine,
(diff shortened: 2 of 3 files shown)


tsloughter:
It looks like metrics could also be improved by having it check if the module exists and then only hit the code server if it didn't find it. Like:

```
erlang:function_exported(Module, module_info, 0) orelse code:ensure_loaded(Module)
```

But it doesn't look worth it to do it as often as it was before this PR anyway, so not as important a change to the metrics app.

benoitc:
hrm the fact that the module is recompiled everytime is an issue... Changes looks good for me.

benoitc:
tough the point of using a metric module was to not use ets... maybe it worth to revisit that completely. Another option is to go for latest version of erlang-metrics. What do you think ?

scohen:
Hey fellas, was there something wrong with this PR? 22 days is a long time to wait.

benoitc:
@scohen vacations and time offline.....

benoitc:
i will revisit that in next version that removes hackney_manager

scohen:
👍

Hinidu:
@scohen Nice improvements! Will try them in my project as soon as they will be released.
Btw, you reminded me about another similar performance issue described here: https://github.com/certifi/erlang-certifi/issues/19. Perhaps you or somebody else will be interested in it too.

Issue #447, opened 2017-10-20
Deadlocks and excessive CPU consumption with async responses.


obmarg:
I've (again) been load testing a proxy application that makes use of async responses, against the latest master branch of hackney.  

I'm basically spinning up 1000 clients that talk to our service, each of the requests those clients make will kick off a request using hackney.  After a reasonably small number of requests (8000-9000 pre `1ab27d4`, _seems_ to be more after that) things start going wrong:

1. One of my processes that was receiving a response crashes.
2. Hackney manager crashes with 

```
13:10:34.417 [error] GenServer :hackney_manager terminating
** (MatchError) no match of right hand side value: {:error, :normal}
    (hackney) /Users/graeme/src/thirdparty/hackney/src/hackney_manager.erl:525: :hackney_manager.clean_requests/5
    (hackney) /Users/graeme/src/thirdparty/hackney/src/hackney_manager.erl:499: :hackney_manager.handle_owner_exit/4
    (stdlib) gen_server.erl:616: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:686: :gen_server.handle_msg/6
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
```

3. Hackney seems to be deadlocked somehow, as no further requests ever complete - though nothing ever crashes either.
3. beam.smp processor usage shoots up to 500%.  This seems to be caused by `hackney_sup` - which permanently has a message queue of around 1000.

I've done some bisecting around this issue, and it definitely seems to have been introduced as part of #443

benoitc:
@obmarg is this with 0.10.1 ? Pb should be fixed in this version . Let me know.

obmarg:
Yes, this is with 0.10.1.  The fix you made in 0.10.1 does seem to have made the error condition less likely, but it still happens sooner or later.

benoitc:
@obmarg can you share a snippet of your code? I would like to understand what it does.

benoitc:
not sure how you get that error on line 525 though... I don't see how it can happen right now.

benoitc:
@obmarg also what is in the supervisor queue? Did you look at it? (erlang:process_info or observer should do the trick)

obmarg:
The code itself is reasonably simple, so sure.  I've got this in one function:

```
HTTPoison.request(
  request.method,
  request.url,
  request.body,
  request.headers,
  hackney: [follow_redirect: false],
  timeout: deadline,
  recv_timeout: deadline,
[...]

PR #473, opened 2018-02-07
ssl_closed fix

AlexanderKai:
Here is a few lines of code that can fix up issue #464 in my opinion.

### DIFF
--- a/src/hackney_stream.erl
+++ b/src/hackney_stream.erl
@@ -113,6 +113,9 @@ maybe_continue(Parent, Owner, Ref, #client{transport=Transport,
       Transport:setopts(Socket, [{active, false}]),
       Transport:controlling_process(Socket, From),
       From ! {Ref, ok};
+    {ssl_closed, _} ->
+      Owner ! {hackney_response, Ref, {status, 500, <<"ssl_closed">>}},
+      hackney_response:close(Client);
     {Ref, close} ->
       hackney_response:close(Client);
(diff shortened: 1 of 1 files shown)


benoitc:
I willl test it later today but looks good so far. Thanks for the patch!

stephenmoloney:
I tried to backport these changes to `1.10.1` but I am still running into the same error.
I can't upgrade to `1.11.0` as I keep running into the error on #447. I can reproduce the error in #447 and I'll post that over on that issue once i get the chance.

The local tests don't seem to output the error but building on travis does. 

`1.10.1` -> https://travis-ci.org/stephenmoloney/fastimage/builds/345600324

`1.10.1 with ssl_fix` -> https://travis-ci.org/stephenmoloney/fastimage/builds/345602341

AlexanderKai:
@stephenmoloney 
That fix don't solves issue #447.

critch:
While testing this for #479 I ran into what might be a worse situation. While not crashing out with the unexpected error message, it will not return the done message. So in effect, the download loop will fail to exit do to never receiving the done message.

AlexanderKai:
@critch 
Can you send me test case for reproduce it?

benoitc:
have a look in #482 as well, that should probably fix the error but let me know.

benoitc:
i commited 91aadac9f929966087eef73fe09d1131441dd129 which should fix the issue.

PR #481, opened 2018-02-27
handle response body when the on HTTP 1.0 like response

### LINKED ISSUE #475: Empty response body when Content-Length and Transfer-Encoding headers are omitted
Beginning with v1.11.0, if the response from the server omits both the `Content-Length` and `Transfer-Encoding` response headers, the body of the response is not parsed and returned.

### v1.11.0 Response Body

screen_shot_2018-02-14_at_10_40_38_am

### v1.10.1 Response Body

screen_shot_2018-02-14_at_10_45_46_am

I believe the bug was introduced with the change in https://github.com/benoitc/hackney/pull/450 to handle empty/invalid content-length.

https://github.com/benoitc/hackney/commit/5b51fb82547dd3a492aaac5d63f7aa1ada1e65cf#diff-03059c69a9df088e2476724af69f06bdL334

benoitc:
this change handle the case where no content-length and
transfer-encoding headers are given in the response. The spec says we
should try to read it.

fix #475

### DIFF
--- a/include/hackney_lib.hrl
+++ b/include/hackney_lib.hrl
@@ -24,7 +24,7 @@
  version                :: {integer(), integer()} | undefined,
  method = <<>>          :: binary(),
  partial_headers = []   :: list(),
- clen = undefined        :: integer() | undefined,
+ clen = undefined        :: integer() | undefined | bad_int,
  te = <<>>              :: binary(),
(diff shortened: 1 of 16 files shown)


benoitc:
superseded by #496

PR #487, opened 2018-03-08
Trivial: Replace no-break space (U+00A0) with ASCII in source files.

villiros:
This is cosmetic, but there are a few instances of nbsp in sources; they were probably added by mistake?

These characters make my Erlide refuse to open these files with an error.


benoitc:
thanks!

PR #493, opened 2018-03-24
ensure hackney:setopts/2 returns ok

kevinwilson541:


benoitc:
can you describe the issue you're solving? In which case didn't it return OK?

kevinwilson541:
Sure thing.

This function call at hackney:setopts/2 https://github.com/benoitc/hackney/blob/master/src/hackney.erl#L108 resolves to this function call here https://github.com/benoitc/hackney/blob/master/src/hackney_manager.erl#L226.  The `Func` parameter there returns the result of hackney_manager:update_state/2, which returns the result of erlang:put/2. This will return the value previously stored in the process dictionary instead of `ok`.

Came across this when trying to update the request options of an existing connection through hackney:send_request/3, which makes an explicit match against `ok` here https://github.com/benoitc/hackney/blob/master/src/hackney.erl#L337.

benoitc:
@kevinwilson541 thanks for the feedback!

PR #494, opened 2018-03-24
Fix accessing HTTPS sites with an IPv4 address

piotrklibert:
Hi, I was writing an internal monitoring tool and found that I need to access URLs in the form of `https://123.123.123.123/something/` - that is, URLs with HTTPS protocol and numeric IP in the host part. Unfortunately, when I tried issuing such requests I discovered that Hackney doesn't allow this case. (Actually, I was writing in Elixir with HTTPoison, which in turn uses Hackney.)

After some time debugging I discovered that Erlang's ssl application supports requests to IP addresses only if they are presented as four-tuples of ints, not as strings.

I created a quick fix for this which works for me - just before the `ssl:connect` I split the Host string on dots and check if they can be interpreted as an IP address. I'm not sure if it's any good, it's fine if it's not, but then I think it would be worth it to at least state in the README that HTTPS+IP doesn't work.

### DIFF
--- a/src/hackney_ssl.erl
+++ b/src/hackney_ssl.erl
@@ -50,7 +50,7 @@ messages(_) -> {ssl, ssl_closed, ssl_error}.
 %% string and either returns such tuple, or the string if it's impossible to
 %% parse.
 parse_address(Host) when is_list(Host) ->
-  case inet:parse_address() of
+  case inet:parse_address(Host) of
     {ok, Address} -> Address;
     {error, _} -> Host
   end.
--- a/src/hackney_ssl.erl
+++ b/src/hackney_ssl.erl
@@ -49,6 +49,13 @@ connect(Host, Port, Opts) ->
 
 connect(Host, Port, Opts, Timeout) when is_list(Host), is_integer(Port),
                                         (Timeout =:= infinity orelse is_integer(Timeout)) ->
+  Host1 = try
+            Parts = string:split(Host, ".", all),
+            4 = length(Parts),
+            list_to_tuple(lists:map(fun list_to_integer/1, Parts))
+          catch
+            _:_ -> Host
+          end,
   BaseOpts = [binary, {active, false}, {packet, raw},
     {secure_renegotiate, true},
     {reuse_sessions, true},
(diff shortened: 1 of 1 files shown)


src/hackney_ssl.erl
```diff
@@ -49,6 +49,13 @@ connect(Host, Port, Opts) ->
 
 connect(Host, Port, Opts, Timeout) when is_list(Host), is_integer(Port),
                                         (Timeout =:= infinity orelse is_integer(Timeout)) ->
+  Host1 = try
```
benoitc: would be better to use  `inet:parse_address/1` there. Can you update your PR to use it? (sorry for the late answer)

piotrklibert: Sure, no problem. Just to be sure, you mean something like this, right?
```
f(Host) ->
    Host1 = case inet:parse_address(Host) of
             {ok, Address} -> Address;
             {error, _} -> Host
         end,
    Host1.
```
Would it be better if I moved this to a separate function, or should it remain inline in `connect/4`?

I will probably update the PR when I get home in the evening (can't really code right now, family&stuff...)

benoitc: maybe in a separate function to make it more readable.

benoitc: @piotrklibert would be cool if you can send your commit today to be part of the release. Let me know :)

piotrklibert:
Yes, it seems to be broken still for the IPv6 addresses (testing with Google's `2001:4860:4860::8888`). I don't immediately see why or where it happens, I will need some time to figure it out. IPv4 works (tested after the changes to PR).

piotrklibert:
Scratch the above, it now kind of works, from what I understand as long as the `no_proxy_sni_support` here: https://github.com/benoitc/hackney/blob/6ef0f9bb2cdfa654643be8324e081bf7f05d659a/src/hackney_http_connect.erl#L27-L37 is defined (that is, when the `service_name_indication` is added to options).

OvermindDL1:
Precisely the same issue here, I keep getting back an `%HTTPoison.Error{id: nil, reason: {:options, {{:server_name_indication, '10.0.1.2'}}`...

OvermindDL1:
Tested other commits, it was broken on 6ef0f9bb2cdfa654643be8324e081bf7f05d659a.

OvermindDL1:
And manually applying this patch did indeed fix it.  :-)

Any chance of a release?

benoitc:
@OvermindDL1 thanks, sorry for the delay to apply it.

OvermindDL1:
Awesome thanks!  Any chance on a hex.pm release so I don't have to depend on a git branch anymore?  ^.^

PR #495, opened 2018-03-31
fix env lost after hot upgrade

### LINKED ISSUE #492: Requests fail after hot code upgrade
Since 98f5ac60b1e67a5fbbb3fa9b002ff699ca429c97 was merged (which moved the `metrics:init` call into start/2), a hot code upgrade of my app results in failed hackney calls until I fully restart my elixir application.

> 2018-03-22 07:07:36.840  [error] Ranch listener ApiWeb.Endpoint.HTTP had connection process started with :cowboy_protocol:start_link/4 at #PID<0.3108.0> exit with reason: {{:badmatch, :undefined}, [{:hackney_connect, :create_connection, 5, [file: '/data/www/pfcs-api-builds/deps/hackney/src/hackney_connect.erl', line: 69]}, {:hackney_connect, :connect, 5, [file: '/data/www/pfcs-api-builds/deps/hackney/src/hackney_connect.erl', line: 37]}, {:hackney, :request, 5, [file: '/data/www/pfcs-api-builds/deps/hackney/src/hackney.erl', line: 315]}, {ExAws.Request.Hackney, :request, 5, [file: 'lib/ex_aws/request/hackney.ex', line: 20]}, {ExAws.Request, :request_and_retry, 7, [file: 'lib/ex_aws/request.ex', line: 38]}, {ExAws.Operation.ExAws.Operation.S3, :perform, 2, [file: 'lib/ex_aws/operation/s3.ex', line: 40]}, {ExAws.S3.Upload, :initialize, 2, [file: 'lib/ex_aws/s3/upload.ex', line: 40]}, {ExAws.Operation.ExAws.S3.Upload, :perform, 2, [file: 'lib/ex_aws/s3/upload.ex', line: 82]}]}

I traced this back to following line in hackney_connect.erl returning `undefined` after the reload:

```
{ok, Engine} = application:get_env(hackney, metrics),
```

I believe this is because the new `set_env` call is reset by the app config in hackney.app.src when it's reloaded, removing the `metrics` option?

redink:
Fixes #492

### DIFF
--- a/src/hackney_connect.erl
+++ b/src/hackney_connect.erl
@@ -66,7 +66,7 @@ create_connection(Transport, Host, Port, Options, Dynamic)
   MaxBody = proplists:get_value(max_body, Options),
 
   %% get mod metrics
-  {ok, Engine} = application:get_env(hackney, metrics),
+  Engine = application:get_env(hackney, metrics, nil),
 
   %% initial state
   InitialState = #client{mod_metrics=Engine,
@@ -279,7 +279,9 @@ check_mod_metrics(#client{mod_metrics=Mod}=State)
   when Mod /= nil, Mod /= undefined ->
   State;
 check_mod_metrics(State) ->
-  State#client{mod_metrics=metrics:init(hackney_util:mod_metrics())}.
+  Metrics = metrics:init(hackney_util:mod_metrics()),
+  application:set_env(hackney, metrics, Metrics),
+  State#client{mod_metrics = Metrics}.
 
 ssl_opts(Host, Options) ->
   case proplists:get_value(ssl_options, Options) of


src/hackney_connect.erl
```diff
@@ -66,7 +66,7 @@ create_connection(Transport, Host, Port, Options, Dynamic)
   MaxBody = proplists:get_value(max_body, Options),
 
   %% get mod metrics
-  {ok, Engine} = application:get_env(hackney, metrics),
+  Engine = application:get_env(hackney, metrics, nil),
```
benoitc: that doesn't sound good, it should be already initialized, if not raises an error

benoitc:
thanks for the patch. I went further though by not rlying at all on the env for it so it wouldn't be lost across upgrade. see 9604828584457724d1502a535edc8cc8b0f43b94

redink:
And, I can't get 0.12 from hex.pm . Please take a look into it. Thanks very much.

benoitc:
0.12 is not yet released. It's planned for later today. However master can be tested....

PR #499, opened 2018-04-03
fix terminate_async_response

### LINKED ISSUE #498: Deadlocks in `hackney_manager:wait_async_response`
I'm seeing deadlocks in `hackney_manager:wait_async_response` with the latest master branch of hackney.

The stack trace of the hackney_manager process is:
image

I can reproduce this fairly consistently in the tests for a couple of private projects, but dont have an easy code sample to reproduce it.

I believe what's happening is:

1. [Hackney monitors the stream process](https://github.com/benoitc/hackney/blob/master/src/hackney_manager.erl#L562)
2. It receives an exit message in monitor_child, and listens for DOWN messages.
2. [It receives a down noproc message from the monitor](https://github.com/benoitc/hackney/blob/master/src/hackney_manager.erl#L549-L550)
3. It enters wait_async_response, to wait for a down message.
4. That message never arrives, because it's already been read in monitor_child.

I've tried just removing the inner receive block from `monitor_child` and this stops the issue from occurring.  However, I assume that block is there for a reason and I'm not sure what other impact removing it has...

benoitc:
following observations of @obmarg delay the monitoring of the stream process in wait_async_response instead of doing it in monitor_child.

fix #498

### DIFF
--- a/src/hackney_manager.erl
+++ b/src/hackney_manager.erl
@@ -540,32 +540,20 @@ clean_requests([], _Pid, _Reason, _PoolHandler, State) ->
 monitor_child(Pid) ->
   erlang:monitor(process, Pid),
   unlink(Pid),
-
   receive
-    {'EXIT', Pid, Reason} ->
-      receive
-        {'DOWN', _, process, Pid, normal} ->
-          ok;
-        {'DOWN', _, process, Pid, noproc} ->
-          ok;
-        {'DOWN', _, process, Pid, _} ->
-          {error, Reason}
-      end
+    {'EXIT', Pid, _} ->
+      true
   after 0 ->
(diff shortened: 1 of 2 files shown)


obmarg:
I've run a whole bunch of tests on this branch, and it seems to fix the deadlock I was experiencing.

Unrelated, but would it be OK to remove [this logging line](https://github.com/benoitc/hackney/blob/master/src/hackney_response.erl#L167).  It's causing quite a lot of pointless output in my tests, and I assume it was just added for debugging.

benoitc:
thanks for the feedback :) i will merge once i’m back from the swimming pool. Will also remove that spurious log line... good catch!

PR #520, opened 2018-08-17
(feat) Support 'checkout_timeout' option

leonardb:
Allow caller to handle overload when pool is maxed out rather than returning
incorrect 'connection_timeout' response


src/hackney_pool.erl
```diff
@@ -67,9 +67,9 @@ checkout(Host0, Port, Transport, #client{options=Opts}=Client) ->
   RequestRef = Client#client.request_ref,
   Name = proplists:get_value(pool, Opts, default),
   Pool = find_pool(Name, Opts),
-  ConnectTimeout = proplists:get_value(connect_timeout, Opts, 8000),
+  CheckoutTimeout = proplists:get_value(checkout_timeout, Opts, 8000),
```
benoitc: this introduce a breaking change. I would rather add for the current version add support for both options, that may have a different meaning later  or be deprecated. Can you do the change?

leonardb: @benoitc As I read it does `support both options` with the change. Unless you're wanting to fall back through the options should `checkout_timeout` not be defined.

IE
`checkout_timeout -> connect_timeout -> default 8000`

benoitc: I meant to fallback for now. ie if connect_timeout is set using it as checkout_timeout.

leonardb: What I was proposing is to fall back to `connect_timeout` when `checkout_timeout` is not set.
If we do it the other way around `checkout_timeout` would not be used when set and the desired separation of connect/checkout would not exist

```diff
diff --git a/src/hackney_pool.erl b/src/hackney_pool.erl
index ceb97a9..599fcf5 100644
--- a/src/hackney_pool.erl
+++ b/src/hackney_pool.erl
@@ -67,7 +67,9 @@ checkout(Host0, Port, Transport, #client{options=Opts}=Client) ->
   RequestRef = Client#client.request_ref,
   Name = proplists:get_value(pool, Opts, default),
   Pool = find_pool(Name, Opts),
-  CheckoutTimeout = proplists:get_value(checkout_timeout, Opts, 8000),
+  ConnectTimeout = proplists:get_value(connect_timeout, Opts, 8000),
+  %% Fall back to using connect_timeout if checkout_timeout is not set
+  CheckoutTimeout = proplists:get_value(checkout_timeout, Opts, ConnectTimeout),
   case catch gen_server:call(Pool, {checkout, {Host, Port, Transport},
     Pid, RequestRef}, CheckoutTimeout) of
     {ok, Socket, Owner} ->
```

benoitc:
Thanks! merged :)

PR #535, opened 2018-09-26
don't encode host for unix scheme

### LINKED ISSUE #534: UNIX domain sockets broken with new idna dependency change
We have some common tests that use `hackney:request` to test both TCP and UNIX domain socket behavior. In the past week, the upgrade of hackney to the 6.0.0 version of benoitc/idna broke the URIs we pass into this request. Either we were broken in how we handled this before, or the hackney library isn't taking named sockets into account with the inda change.

In our specific test, our socket name is `/tmp/foobar.sock` which we then pass to `hackney_url:urlencode` and then build a full URI out of it with the `http+unix` protocol resulting in a URI of:
```erlang
<<"http+unix://%2ftmp%2ffoobar.sock/foobar/v1/foobar_test?requestid=F07084CC-7628-1005-ABE3-A45E60C8DA2B">>
```
The error we receive then this:
```erlang
%%% foobar_SUITE ==> {bad_label,{alabel,"The label \"/tmp/foobar\"  is not a valid A-label: ulabel error={bad_label,\n                                                                {context,\n                                                                 \"Codepoint 47 not allowed ('DISALLOWED') at posion 0 in \\\"/tmp/foobar\\\"\"}}"}}
```

If we pin our build to idna 5.1.2, we do not receive the above error.

benoitc:
fix #534

### DIFF
--- a/src/hackney_url.erl
+++ b/src/hackney_url.erl
@@ -90,10 +90,14 @@ normalize(#hackney_url{}=Url, Fun) when is_function(Fun, 1) ->
                        {Host0, Netloc0};
                      _ ->
                        Host1 = unicode:characters_to_list(
-                         urldecode(unicode:characters_to_binary(Host0))),
+                                 urldecode(unicode:characters_to_binary(Host0))
+                                ),
 
                        %% encode domain if needed
-                       Host2 = idna:to_ascii(Host1),
+                       Host2 = case Scheme of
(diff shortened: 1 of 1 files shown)


zp-sd:
@benoitc This one is still crashing - https://github.com/benoitc/hackney/blob/master/src/hackney_connect.erl#L39.

PR #542, opened 2018-10-30
Fix doc for checkout_timeout default

Hinidu:
Hello!
Current doc is a bit misleading because if you have custom `connect_timeout` when default `checkout_timeout` won't be equal to 8000.


jszwedko:
:+1: this one confused me initially too. It wasn't until I looked at the source that I realized it defaults to the `connect_timeout`.

PR #543, opened 2018-10-30
Replace cacerts usage by cacertfile

Hinidu:
Hello!
Finally I found a little bit of time to make this PR after https://github.com/certifi/erlang-certifi/issues/19 have been merged.
It gives quite big performance improvements and it's cool that it will work out of the box for everyone :-)

### DIFF
--- a/src/hackney_connect.erl
+++ b/src/hackney_connect.erl
@@ -323,7 +323,6 @@ ssl_opts(Host, Options) ->
 
 ssl_opts_1(Host, Options) ->
   Insecure =  proplists:get_value(insecure, Options, false),
-  CACerts = certifi:cacerts(),
   case Insecure of
     true ->
       [{verify, verify_none}];
@@ -334,7 +333,7 @@ ssl_opts_1(Host, Options) ->
        },
       [{verify, verify_peer},
        {depth, 99},
-       {cacerts, CACerts},
+       {cacertfile, certifi:cacertfile()},
        {partial_chain, fun partial_chain/1},
        {verify_fun, VerifyFun}]
   end.


benoitc:
what would it give a performance improvements though?

Hinidu:
OTP does caching with `cacertfile` but parse `cacerts` on every request. The more detailed analysis can be found here: https://github.com/certifi/erlang-certifi/issues/19
I think that almost any possible workload/benchmark would benefit from this change. Of course it should be checked, I will try to send my benchmarks soon.

Hinidu:
Measurements tip: improvements are more visible if pool is disabled.

benoitc:
The patch looks good for me. The issue is that some libraries, especially those who are working on escript are relying on the current behaviour so it can't be a minor change. Maybe the result of cacerts() can also be cached?

Hinidu:
AFAIR it can’t be effectively cached without changes in OTP ssl API.

Just curious why it won’t work with escript? escript has no access to priv directory?

By the way It would be fine enough if I could specify usage of cacertfile in my config but I can’t call a third party library function in old fashioned Elixir config. I’m gonna try distillery 2 and I hope that it’s config providers will allow me to do that.

benoitc:
If the file live in an external folder you would have to ship the file with it or provide a location. In any case we can't break the current usage, so it can't be a minor update. 

iIt's possible to feed the cache (dist or local) with the content of cacert using the `ssl_pem_cache` module.  I'm not sure yet what is the difference as I though it was just caching the decoded pems and nothing much though.

benoitc:
looking at that patch, it can be clearly useful. However an option should be given to not use that file for applications that need it. i will revisit it to handle thios case. Taking it open until it's done.

benoitc:
fixed by fixed by 7cbe1f6f9a2c464a5ec4dcccca0a7ed41005b9af while I was there. Thanks @Hinidu for the suggestion!

Hinidu:
Hello @benoitc!

Thank you for using my suggestion. But why did you lose that change on the next day here? https://github.com/benoitc/hackney/commit/314b3dbd3a7f8174336062e4948216e88b5ac917 Did you do that by mistake or by purpose?

Thank you in advance.

PR #546, opened 2018-12-19
multipart: DOM element name definition for files

m-2k:
Sending form with a field names for files

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -236,6 +236,7 @@ the request body:
   - `eof`: end the multipart request
   - `{file, Path}`: to stream a file
   - `{file, Path, ExtraHeaders}`: to stream a file
+  - `{file, Path, Name, ExtraHeaders}` : to stream a file with DOM element name
   - `{Name, Content}`: to send a full part
   - `{Name, Content, ExtraHeaders}`: to send a full part
   - `{mp_mixed, Name, MixedBoundary}`: To notify we start a part with a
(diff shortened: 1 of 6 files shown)


src/hackney_multipart.erl
```diff
@@ -67,9 +67,16 @@ encode_form(Parts, Boundary) ->
                     {ok, Bin} = file:read_file(Path),
                     PartBin = << MpHeader/binary, Bin/binary, "\r\n"  >>,
                     {AccSize1, << AccBin/binary, PartBin/binary >>};
-                ({file, Path, Disposition, ExtraHeaders}, {
+                ({file, Path, <<Name/binary>>, ExtraHeaders}, {AccSize, AccBin}) ->
```
benoitc: i would rather match it after the `{file, Path, {Disposition, Params}, ..}` clause since it will match a tuple. Also use the is_binary guard not `<<Name/binary>>`

doc/README.md
```diff
@@ -236,6 +236,7 @@ the request body:
   - `eof`: end the multipart request
   - `{file, Path}`: to stream a file
   - `{file, Path, ExtraHeaders}`: to stream a file
+  - `{file, Path, Name, ExtraHeaders}` : to stream a file with DOM element name
```
benoitc: this change should be done in the `overview.edoc` file. the README is generated from it.

src/hackney_multipart.erl
m-2k: fixed

doc/README.md
m-2k: fixed

src/hackney_multipart.erl
benoitc: thanks!

Issue #550, opened 2019-01-08
Getting 500 response for valid url


idod:

benoitc:
can you provide more informations: logs, trace or anything that could help to reproduce the issue?

idod:
Sorry, 
I got an error on hackney and not on curl.
No I get it also on curl :)
No issue.

idod:
So there is an issue:
when trying to get the url of "https://assets.superbalistcdn.co.za/500x720/filters:quality(75):format(jpg)/spree/product/s/p/spree2018september12picslgzh0650.jpg" with hacnkey I get an error, although it is working fine on curl / browser:

when getting with hackney:
hackney:get(<<"https://assets.superbalistcdn.co.za/500x720/filters:quality(75):format(jpg)/spree/product/s/p/spree2018september12picslgzh0650.jpg">>).
{ok,500,
    [{<<"Date">>,<<"Tue, 08 Jan 2019 14:11:53 GMT">>},
     {<<"Content-Type">>,<<"text/html; charset=UTF-8">>},
     {<<"Transfer-Encoding">>,<<"chunked">>},
     {<<"Connection">>,<<"keep-alive">>},
     {<<"Set-Cookie">>,
      <<"__cfduid=d8081ddfb7dd7ce93d45546477584bcc11546956713; expires=Wed, 08-Jan-20"...>>},
     {<<"X-Varnish">>,<<"15930532">>},
     {<<"Via">>,<<"1.1 varnish-v4, 1.1 google">>},
     {<<"Alt-Svc">>,<<"clear">>},
     {<<"CF-Cache-Status">>,<<"MISS">>},
     {<<"X-Content-Type-Options">>,<<"nosniff">>},
     {<<"Expect-CT">>,
      <<"max-age=604800, report-uri=\"https://report-uri.cloud"...>>},
     {<<"Server">>,<<"cloudflare">>},
     {<<"CF-RAY">>,<<"495f41806f2029dd-SEA">>}],
    #Ref<0.2804061591.1561329665.145241>}

When curl:
curl -v "https://assets.superbalistcdn.co.za/500x720/filters:quality(75):format(jpg)/spree/product/s/p/spree2018september12picslgzh0650.jpg" -O /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
[...]

PR #551, opened 2019-01-09
fixed pathencode to support '(' ')' characters

idod:
Related to  #550 

After checking the [RFC ](https://tools.ietf.org/html/rfc3986#section-2.2) section 2.2, Reserved Characters, sub-delims, it seems that url path with '(' ')' characters is valid and should not be encoded.
I would like to do the same for the other characters as well, and do a small refactor that will reflect the RFC.
What do you think?

Cheers

### DIFF
--- a/src/hackney.app.src
+++ b/src/hackney.app.src
@@ -4,7 +4,7 @@
 {application, hackney,
   [
     {description, "simple HTTP client"},
-    {vsn, "1.15.1"},
+    {vsn, "1.15.0"},
     {registered, [hackney_pool]},
     {applications, [kernel,
       stdlib,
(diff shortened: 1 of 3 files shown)


benoitc:
similar to #251

src/hackney.app.src
```diff
@@ -4,7 +4,7 @@
 {application, hackney,
   [
     {description, "simple HTTP client"},
-    {vsn, "1.15.0"},
+    {vsn, "1.15.1"},
```
benoitc: can you make  remove that change? I will take care of the release numbering.

idod: Done

benoitc:
@idod thanks!

PR #553, opened 2019-01-30
Support empty body as list for POST and PUT operations

miavisa:


miavisa:
@benoitc please can you reviews this? Currently I am using `swagger-codegen` to generate an erlang-client, it uses as dependency `hackney`.

The client handle the empty bodies as an empty list `[]`. From my point of view use a empty list as an empty body is correct, and `hackney` should support this for `POST` and `PUT` operations like the case of empty binary `<<>>`.

benoitc:
@miavisa just merged it. That case was missing indeed. Thanks!

PR #555, opened 2019-02-13
Clarify what recv_timeout means

neenaoffline:
Hi.

The documentation for recv_timeout is misleading. It may be confused as timeout for fetching a connection from a connection pool. Can we make a small change to clarify this?


benoitc:
i agreed. just merged it. thanks!

PR #561, opened 2019-03-24
correct the spec for stop_async/1

rciorba:
Hello! 

The spec for `stop_async` seems wrong. The function returns `{ok, client_ref()}` on success not `ok`.
This seems to be confirmed by `examples/test_async_once2.erl`.

```erlang
{ok, Ref} = hackney:stop_async(Ref),
```

Regards,
Radu

P.S. Thanks for providing this library!


benoitc:
thanks!

PR #562, opened 2019-04-24
store/3 creates invalid dict when called with list of Values

Bencepapa:
store/3 creates invalid dict when called with list of Values, causing function_clause in to_list


benoitc:
thanks!

PR #575, opened 2019-06-04
bump ssl_verify_fun to 1.1.5

aboroska:
To get rid of the `===> ... ssl_verify_fun.app" is missing description entry` warning produced by the latest `rebar3`.


chulkilee:
I'm wondering whether hackney should pin down all package versions. Is there any reason we don't use `~>` for deps?

benoitc:
@chulkilee main reason is that you can't trust a lib is really doing what is advertised by this version number. that is just that :)

PR #576, opened 2019-06-12
Fixes test run example in the README

adkron:
Originally the command for running the tests was `make test`. There is no `make
test` task, but there is `rebar3 eunit` which is mentioned higher in the README.
I tested the tests replacing `make test` with `rebar3 eunit` and everything ran
so I assume this is the correct way of running the tests.

Amos King @adkron <amos@binarynoggin.com>


benoitc:
this change should be part of the overview.edoc from which the README is generated. Can you do the change there?

PR #579, opened 2019-06-21
improve docs and type spec

woylie:
Neither doc nor type spec listed the return type when `with_body` is set to true.


benoitc:
thanks!

PR #584, opened 2019-08-13
Failed tests without any commits on a forked repository

techno-tanoC:
I forked the repository, and run tests on Travis CI.

Although I didn't commit, some tests failed.

I watched logs and noticed that CI fetches the `ranch` version `2.0.0-rc.1`. (the original tests fetched the `ranch` version `1.7.1`)

https://travis-ci.org/techno-tanoC/hackney/jobs/568343040#L515
https://travis-ci.org/benoitc/hackney/jobs/563376913#L514

By the way, I updated the `cowboy` to version `1.1.2` , and Travis CI fetched the `ranch` version `1.3.2`. All tests passed as a result.

https://travis-ci.org/techno-tanoC/hackney/jobs/571190015#L515

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: erlang
+dist: trusty
 otp_release:
   - 21.0.2
   - 20.3
--- a/rebar.config
+++ b/rebar.config
@@ -35,7 +35,7 @@
                            {top_level_readme,
                             {"./README.md", "http://github.com/benoitc/hackney"}}]}]},
              {test, [
-               {deps, [{cowboy, "1.0.4"}, {jsone, "1.4.7"}]}
+               {deps, [{cowboy, "1.1.2"}, {jsone, "1.4.7"}]}
              ]}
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: erlang
+dist: trusty
```
benoitc: part of the patch have already been merged via another one but i'm curious about that change. Why changing the distribution?

benoitc: ?

techno-tanoC: Looking the [Travis's document](https://docs.travis-ci.com/user/languages/erlang/), Travis doesn't support OTP 20.3 on Xenial or later. So to run tests on erlang-20.3, we should use trusty.

Now, hackney is not tested on erlang-20.3. It's meaningless to change the distribution.

benoitc:
sorry it took to much tome to handle it. We are now testing only against last versions.

PR #588, opened 2019-09-18
Remove honor_cipher_order SSL option

nifoc:
Setting `honor_cipher_order` returns the following error when used with
OTP 22.1: `{option, server_only, honor_cipher_order}`

It seems like older versions simply ignored the option, since even the
original PR mentions it being an option for the server.


nifoc:
All test failures seem to be unrelated.

benoitc:
Thanks for the patch. I merged #590 that added 22 to the CI but thanks again!

PR #589, opened 2019-09-18
Fix error caused by `honor_cipher_order`, CI errors

jonnystorm:
OTP-22.1 (commit `124da8950c`) introduced `assert_role/4` into `:ssl`, which throws an error when `:ssl.connect` is called with the `honor_cipher_order` option defined. The error returned is `{:error, {:option, :server_only, :honor_cipher_order}}`.

With OTP-22.1, a dialyzer warning also appears:

```
    129: Invalid type specification for function
    hackney_ssl:shutdown/2. The success typing is
    ( { 'sslsocket',
         _,
         nonempty_maybe_improper_list()
      | {'dtls' | port(), _}
      },
      'read' | 'read_write' | 'write'
    ) -> 'ok' | {'error',_}
```

Prior to the above, CI errors began to appear as a result of
cowboy-1.0.4 download failures. Using cowboy-1.1.2 fixes
this.

* remove `honor_cipher_order` option in `connect/4`
* add OTP-22.0 to .travis.yml
* drop OTP-20.2 from .travis.yml (no longer available)
* fix invalid type spec for `shutdown/2`
* require cowboy-1.1.2

### DIFF
--- a/src/hackney_ssl.erl
+++ b/src/hackney_ssl.erl
@@ -65,7 +65,6 @@ connect(Host, Port, Opts, Timeout) when is_list(Host), is_integer(Port),
   BaseOpts = [binary, {active, false}, {packet, raw},
     {secure_renegotiate, true},
     {reuse_sessions, true},
-    {honor_cipher_order, true},
     {versions,['tlsv1.2', 'tlsv1.1', tlsv1, sslv3]},
     {ciphers, ciphers()}],
   Opts1 = hackney_util:merge_opts(BaseOpts, Opts),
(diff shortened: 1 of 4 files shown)


src/hackney_ssl.erl
```diff
@@ -126,7 +125,7 @@ close(Socket) ->
 
 %% @doc Immediately close a socket in one or two directions.
 %% @see ssl:shutdown/2
--spec shutdown(ssl:socket(), read | write | read_write) -> ok.
+-spec shutdown(ssl:sslsocket(), read | write | read_write) -> ok | {error, any()}.
```
benoitc: hrm i wonder how the spec works with 19.3 but I will check

PR #590, opened 2019-09-19
Remove setting honor_cipher_order client side

vihu:
This option is not  meaningful to set on the client side, it worked prior to the check added in OTP22.1 here: 
https://github.com/erlang/otp/commit/124da8950c1fe93ef02cb044157e78798bc2360d#diff-79b520311347e34fac9cffbc4ea373f9R1720


benoitc:
why closing?

vihu:
#588 and #589 seem to be aiming for the same error I noticed today. I suppose OTP22.1 _really_ highlighted this problem.

benoitc:
noticed it after , yes. I will take care about the issue sometimes today :)

PR #600, opened 2019-11-27
add test case for broken https proxying

RodyaPonomaryov:
when mix proxy pool and https request sometimes request go to wrong server. this very bad. i create test show issue.

please notify user because setup could hide error and secret data leaked


benoitc:
thanks! That will help for the fix :)

benoitc:
Closing as outdated. This test uses Cowboy 1.x API which is no longer supported. The hackney test infrastructure has been updated to use Cowboy 2.x. Thank you for the contribution!

Issue #566, opened 2019-05-16
Connection mixed up when using proxy


thang911pk:
I experienced similar issue as https://github.com/benoitc/hackney/issues/545
I use the latest version of HTTPoison and Hackney. I had 2 end points. When I use Squid proxy to send HTTPS requests, all requests go to one endpoint. Hackney reused the same connection for both end points. 
```
## Without proxy 
HTTPoison.post!(
  "https://b61fad2e.ngrok.io",
  "hello",
  [],
  [
    hackney: [pool: :test_pool]
  ]
)

HTTPoison.post!(
  "https://b61fad2e.ngrok.io",
  "hello",
  [],
  [
    hackney: [pool: :test_pool]
  ]
)


HTTPoison.post!(
  "https://2ffcbfd7.ngrok.io",
  "hello",
  [],
  [
    hackney: [pool: :test_pool]
  ]
)

HTTPoison.post!(
  "https://b61fad2e.ngrok.io",
  "hello",
  [],
  [
    hackney: [pool: :test_pool]
  ]
)

#Socket connections
:sys.get_state(:hackney_pool.find_pool(:test_pool)) 

 {{[
      [
        {'b61fad2e.ngrok.io', 443, :hackney_ssl},
        {:sslsocket, {:gen_tcp, #Port<0.18470>, :tls_connection, :undefined},
         #PID<0.503.0>}
      ]
    ], [], [], [], [], [], [], [], [], [], [], [], [], [], [],
    [
      [
        {'2ffcbfd7.ngrok.io', 443, :hackney_ssl},
        {:sslsocket, {:gen_tcp, #Port<0.19373>, :tls_connection, :undefined},
         #PID<0.523.0>}
      ]
    ]}}},
 {:dict, 2, 16, 16, 8, 80, 48,
  {[], [], [], [], [], [], [], [], [], [], [], [], [], [], [], []},
  {{[], [], [], [], [], [],
    [
      [
        {:sslsocket, {:gen_tcp, #Port<0.19373>, :tls_connection, :undefined},
         #PID<0.523.0>} |
        {{'2ffcbfd7.ngrok.io', 443, :hackney_ssl},
         #Reference<0.931103709.3369074689.153627>}
      ]
    ],
    [
      [
        {:sslsocket, {:gen_tcp, #Port<0.18470>, :tls_connection, :undefined},
         #PID<0.503.0>} |
        {{'b61fad2e.ngrok.io', 443, :hackney_ssl},
[...]

Issue #355, opened 2016-10-27
pool cannot be used along with http proxy


ananthakumaran:
when pooling is used along with http proxy, hackney reuses the wrong connection and sends the request to different domain than intended.

Example

make request to s3
make request to mailgun

For the second request, instead of creating a new connection to mailgun, hackney reuses the connection to s3 and sends the payload there.

I have attached the trace of hackney_pool:checkout with and without proxy. The connections seem to be indexed by proxy's host/port rather than the real destination's host/port

```
without proxy
=============

(<0.70.0>) call hackney_pool:checkout("s3-ap-southeast-1.amazonaws.com",443,hackney_ssl_transport,{client,{1477,565345,811889},
        {metrics_ng,metrics_dummy},
        hackney_ssl_transport,"s3-ap-southeast-1.amazonaws.com",443,
        <<"s3-ap-southeast-1.amazonaws.com">>,
        [with_body,{recv_timeout,30000}],
        nil,nil,#Ref<0.0.1.1011>,true,hackney_pool,30000,false,5,false,5,nil,
        nil,nil,undefined,start,nil,normal,false,false,true,undefined,false,
        nil,waiting,nil,4096,<<>>,[],undefined,nil,nil,nil,nil,undefined,nil})
(<0.70.0>) returned from hackney_pool:checkout/4 -> {error,no_socket,
                                                     {default,
                                                      #Ref<0.0.1.1011>,
                                                      {"s3-ap-southeast-1.amazonaws.com",
                                                       443,
                                                       hackney_ssl_transport},
                                                      <0.618.0>,
                                                      hackney_ssl_transport}}
(<0.70.0>) call hackney_pool:checkout("s3-ap-southeast-1.amazonaws.com",443,hackney_ssl_transport,{client,{1477,565347,234478},
        {metrics_ng,metrics_dummy},
        hackney_ssl_transport,"s3-ap-southeast-1.amazonaws.com",443,
        <<"s3-ap-southeast-1.amazonaws.com">>,
        [with_body,{recv_timeout,30000}],
        nil,nil,#Ref<0.0.2.4718>,true,hackney_pool,30000,false,5,false,5,nil,
        nil,nil,undefined,start,nil,normal,false,false,true,undefined,false,
        nil,waiting,nil,4096,<<>>,[],undefined,nil,nil,nil,nil,undefined,nil})
(<0.70.0>) returned from hackney_pool:checkout/4 -> {ok,
                                                     {default,
[...]

PR #606, opened 2020-01-15
Use destination as pool reference for hackney_http_connect transport

ennocramer:
When the HTTP CONNECT method is used for proxied requests, sockets are
bound to the destination host and cannot be reused for requests to
other hosts.  The sockets should therefore be index based on the
destination host, not the proxy host.

This setup is still broken when multiple proxies, or proxied and non-proxied connections, are used at the same time, as connections over either proxy are considered identical.

This change should improve the situation for issues #355 and #566

### DIFF
--- a/src/hackney_pool.erl
+++ b/src/hackney_pool.erl
@@ -69,20 +69,30 @@ checkout(Host0, Port, Transport, #client{options=Opts}=Client) ->
   ConnectTimeout = proplists:get_value(connect_timeout, Opts, 8000),
   %% Fall back to using connect_timeout if checkout_timeout is not set
   CheckoutTimeout = proplists:get_value(checkout_timeout, Opts, ConnectTimeout),
-  case catch gen_server:call(Pool, {checkout, {Host, Port, Transport},
+  case Transport of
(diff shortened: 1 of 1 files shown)


ennocramer:
I made a mistake in my tests. This PR does not work as it is supposed to. Please do not merge yet.

ennocramer:
Apologies for the delay with fixing this PR.

I had initially believed it would be easiest to retain the shape of the pool checkin reference and simply swap out the proxy information with the final host. I did not realize that various other functions require access to the proxy transport value and would require passing additional arguments. I now believe the correct fix is to include both the proxy and final host in the checkin reference. This also means that the limitation in my first comment about connections to the same host with different proxy settings no longer applies.

benoitc:
@ennocramer appologies for the delay to answer, the pr shows the right path anyway. I will take it around. IMO proxies could be handled as an independant resource you link o when making a connection either as a process but lso supervision. I need to think a little bit about it.

benoitc:
it's now fixed in master. Thanks for your patch it was a good basis.

PR #611, opened 2020-02-29
Handle Connection: close response header for streams

Kuroneer:
Due to stream's client not storing the Headers, hackney_response:maybe_close ignores the "Connection: close" response header. This causes that a soon-to-be-closed socket is put back in the pool, potentionally being delivered to other client.

With these changes, hackney_response:maybe_close handles the way hackney_stream stores the connection value

### DIFF
--- a/test/hackney_stream_tests.erl
+++ b/test/hackney_stream_tests.erl
@@ -2,10 +2,13 @@
 -include_lib("eunit/include/eunit.hrl").
 -include("hackney_lib.hrl").
 
+-export([checkout/4]).
+
 %% This seems necessary to list the tests including the generator
 all_tests() ->
    [
-    async_request()
+    async_request(),
+    heed_connection_close()
    ].
 
 stream_test_() ->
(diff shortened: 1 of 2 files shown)


test/hackney_stream_tests.erl
```diff
@@ -53,4 +76,5 @@ receive_response(Ref, Dict0) ->
         {hackney_response, Ref, Bin} ->
             Dict1 = orddict:append(body, Bin, Dict0),
             receive_response(Ref, Dict1)
+    after 5000 -> {error, timeout}
```
benoitc: i think i would make it configurable.

test/hackney_stream_tests.erl
```diff
@@ -41,6 +44,26 @@ receive_response(Ref) ->
     StatusCode = orddict:fetch(status, Dict),
     {StatusCode, Keys}.
 
+checkout(_Host, _Port, _Transport, _Client) -> {error, no_socket, make_ref()}.
+heed_connection_close() ->
```
benoitc: typo? shouldn’t it be « head » ?

Kuroneer: I'll change it to "handle"

test/hackney_stream_tests.erl
Kuroneer: Added a parameter to make it configurable by test case

PR #613, opened 2020-03-10
Do not normalize when converting relative redirect URL to absolute

rinpatch:
According to [HTTP/1.1 RFC](https://tools.ietf.org/html/rfc7231#section-7.1.2) the`Location` header value is a single URI-reference. When the reference is relative it should be merged by rules of ["Reference resolution" section of URI RFC](https://tools.ietf.org/html/rfc3986#section-5). From my understanding of that section, no normalization on the relative URI should take place.

### The specific bug I've encountered that this fixes:

Our application generates preview cards for links users submit. We've noticed that the preview links generated for shortened flickr links (i.e `https://flic.kr/p/231oyNG`) are of the wrong picture.

On further investigation I've uncovered that what was happening is:
1. Application uses hackney to request `https://flic.kr/p/231oyNG` with`follow_redirect` set to `true`.
2. Hackney gets the first redirect to`https://www.flickr.com/photo.gne?short=231oyNG`.
3. After requesting that it gets a redirect to `/photo.gne?rb=1&short=231oyNG`
4. Hackney converts that to an absolute url as `https://www.flickr.com/photo.gne%3frb=1%26short=231oyNG`.
5. Hackney requests that url and for some reason that causes flickr to downcase the case-sensitive id of the picture (`231oyng` instead of `231oyNG`), therefore returning the wrong picture.

### DIFF
--- a/src/hackney.erl
+++ b/src/hackney.erl
@@ -884,13 +884,12 @@ absolute_url(RelativeUrl, #client{transport=T, host=Host, port=Port,
                   _       -> <<Path/binary, "/", RelativeUrl/binary>>
                 end
             end,
-  Parsed = hackney_url:normalize(#hackney_url{scheme=Scheme,
-                                              host=Host,
-                                              port=Port,
-                                              netloc=Netloc,
-                                              path=NewPath}),
-  hackney_url:unparse_url(Parsed).
 
+  hackney_url:unparse_url(#hackney_url{scheme=Scheme,
+                                       host=Host,
+                                       port=Port,
+                                       netloc=Netloc,
+                                       path=NewPath}).
 
 %% handle send response
 reply({ok, Data, NState}, _State) ->


rinpatch:
@benoitc ping

benoitc:
patch is ok, sorry for the delay to merge it. I expect to make a release in the week.

feld:
> patch is ok, sorry for the delay to merge it. I expect to make a release in the week.

any update on a release including this fix?

benoitc:
@feld I have one fix to land before it can go but it should be good for a release on thursday this week.

feld:
Hello, just checking in again. I know life is crazy these days.

benoitc:
yeah a little late. The fix will land tomorrow however so expect a release soon now :)

feld:
👋  any news on the release?

benoitc:
hrm time is going too fast. Will take care this week, maybe tomorrow as a minor release.

PR #626, opened 2020-05-21
Fix further issues with TLS/SSL

g-andrade:
Following your recent changes for OTP 23 support in 66415cf63a, I found some other problems which this PRs proposes fixing.

- broken compilation due to failing `ct_expand` parse transform, fixed in 92c0f4c
- crashing HTTPS requests in OTP 19, fixed in 8eddf7c
- occasional TLS negotiation errors in OTP 20, fixed in 533889d

Cheers,

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -5,7 +5,8 @@
             {platform_define, "R14", no_callback_support},
             {platform_define, "^[0-9]+", namespaced_types},
             {platform_define, "^R", no_proxy_sni_support},
-            {platform_define, "^(19)|(20)", no_customize_hostname_check}
+            {platform_define, "^(19)|(20)", no_customize_hostname_check},
+            {platform_define, "^20", buggy_chacha_ciphers}
            ]}.
 
 {erl_first_files,
--- a/rebar.config
+++ b/rebar.config
@@ -8,6 +8,10 @@
             {platform_define, "^20", no_customize_hostname_check}
            ]}.
 
+{erl_first_files,
+ ["src/hackney_ssl_certificate.erl" % required by `hackney_ssl' at compile time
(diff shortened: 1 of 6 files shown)


rebar.config
```diff
@@ -20,6 +25,7 @@
         {mimerl, "~>1.1"},
         {certifi, "2.5.2"},
         {metrics, "1.0.1"},
+        {parse_trans, "3.3.0"},
```
benoitc, line 28: Is this needed? parse_trans is coming with certifi normally

src/hackney_ssl.erl
```diff
@@ -88,16 +109,14 @@ decoded_cacerts() ->
   ct_expand:term(
     lists:foldl(fun(Cert, Acc) ->
                     Dec = public_key:pkix_decode_cert(Cert, otp),
-                    [extract_public_key_info(Dec) | Acc]
+                    [hackney_ssl_certificate:public_key_info(Dec) | Acc]
                 end, [], certifi:cacerts())
    ).
 
 
-extract_public_key_info(Cert) ->
```
benoitc, line 96: It's fine but curious why moving it to another module?

rebar.config
benoitc: it's fine i will commit with it

benoitc:
thanks for the patch!

src/hackney_ssl.erl
benoitc: sorry found the explanation in the commit.... :)

rebar.config
g-andrade: I just meant to be explicit, since `hackney` started depending directly on it.

Issue #636, opened 2020-06-08
TLS support may be broken on OTP 20.0 / 20.1


g-andrade:
As per @okeuday's [comment](https://github.com/benoitc/hackney/commit/533889df19eb51d4600f383bbeaad17bc6bc4fe7#r39625597) on `hackney_ssl.erl`, which correlates with a few errors I saw when running the test suites on multiple OTP versions a few days ago but didn't have the chance to investigate.

Creating this issue mostly as a reminder to fix this issue since I'm the one to blame[*], but also for transparency in case anyone else bumps into the same problem.

[*]: *mea culpa, mea culpa, mea maxima culpa*

g-andrade:
Closed by #637 .

okeuday:
@g-andrade You probably want to use versions from the list https://docs.travis-ci.com/user/languages/erlang/#otprelease-versions for #639 because the Erlang/OTP team has been moving away from release tarballs and using git tags instead (so the "stable" or latest version becomes more of a moving target).

PR #637, opened 2020-06-09
Fix Erlang/OTP 20.0 / 20.1 compilation

okeuday:
* Proposed fix for #636
* Bumped into the issue with the Ubuntu 18.04 Erlang/OTP package using 20.2.
  Travis doesn't appear to have 20.2 but does have 20.1.7
  (https://docs.travis-ci.com/user/languages/erlang/#otprelease-versions).


benoitc:
thanks!

PR #639, opened 2020-06-12
Lower likelihood of regressions

g-andrade:
Make them easier to find early by expanding the range of distinct OTP version targets when running tests on Travis CI.

(In order to prevent issues like #636.)


g-andrade:
Merge conflicts fixed.

PR #640, opened 2020-06-12
Fix erroneous unexpted message

rozap:
* mutually recursive maybe_continue and async_recv
  functions are really tricky to reason about, but
  it doesn't seem like it makes sense for maybe_continue
  to do error handling at all, since async_recv was already
  doing it
  * maybe_continue and stream_loop shouldn't select
    socket messages
  * we're getting into a state where the socket sends
    us a message while we're doing other things and
    we do that receive and then crash
* this change makes async_recv the only one to do
  a select for transport messages

* When the server closes the connection
  and :ssl sends us :ssl_closed, we weren't
  handling the message, which caused it to leak
  into the mailbox even after the request was ended
  * need to flush to make sure it's gone in this case
    where the server hangs up
  * caused when Connection: close header is sent

fix debugging statement

### DIFF
--- a/src/hackney_response.erl
+++ b/src/hackney_response.erl
@@ -349,4 +349,16 @@ close(#client{socket=nil}=Client) ->
   Client#client{state = closed};
 close(#client{transport=Transport, socket=Skt}=Client) ->
   Transport:close(Skt),
+  flush(Transport, Skt),
   Client#client{state = closed, socket=nil}.
+
+
+flush(Transport, Socket) ->
+  {Msg, MsgClosed, MsgError} = Transport:messages(Socket),
+  receive
+    {Msg, Socket, _} -> flush(Transport, Socket) ;
+    {MsgClosed, Socket} -> flush(Transport, Socket) ;
+    {MsgError, Socket, _} -> flush(Transport, Socket)
+  after 0 ->
+    ok
+  end.
--- a/src/hackney_stream.erl
+++ b/src/hackney_stream.erl
@@ -120,11 +120,7 @@ maybe_continue(Parent, Owner, Ref, #client{transport=Transport,
       exit({owner_down, Owner, Reason});
     {system, From, Request} ->
       sys:handle_system_msg(Request, From, Parent, ?MODULE, [],
-        {stream_loop, Parent, Owner, Ref, Client});
-    Else ->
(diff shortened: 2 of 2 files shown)


erik-mocny:
@benoitc Hi, please why is it still not merged as a whole? I can see only second commit changes in https://github.com/benoitc/hackney/compare/1.16.0...v1.17.0 probably cherry-picked? Thanks

benoitc:
i didn't cherry-picked :) I merged your other PR #641 . I didn't merge this one yet as i wanted to test it first. There will be a new release next week.

erik-mocny:
Thanks for the reply. It wasn't my PR I just asked about it because we were facing same issue in our projects.

Dvogiatz:
Hello @benoitc, what is the current status on this PR? 

> There will be a new release next week that should fix the issue.

Also is this referring to a 1.17.1 version coming soon? Thank you

Dvogiatz:
Update: after 2 days of working properly without any issues, it started to throw the same error again.

benoitc:
Sorry to have been silent. There is a new release coming next week that
will fix this issue. I will create a ticket summarizing it.

luizmiranda7:
@benoitc I'm trying the new version of hackney: 1.17.4 but it keeps the same behavior:

```
{:ssl_closed,
 {:sslsocket, {:gen_tcp, #Port<0.330>, :tls_connection, :undefined}
```

- Erlang/OTP 22 [erts-10.7.2.7]
- Hackney 1.17.4

ringofhealth:
```
 [error] Unexpected message: {:ssl_closed, {:sslsocket, {:gen_tcp, #Port<0.213>, :tls_connection, :undefined}, [#PID<0.2999.0>, #PID<0.2998.0>]}}

```

same issue here as well

Erlang/OTP 22 [erts-10.7.2.7]
Hackney 1.17.4

benoitc:
Closing as outdated. The hackney codebase has been significantly refactored to use a gen_statem-based connection model (`hackney_conn.erl`). The files `hackney_response.erl` and `hackney_stream.erl` no longer exist.

Additionally, PR #800 (merged) addresses the ssl_closed detection issue by using `{active, once}` mode on idle pooled connections to properly detect server-initiated closes. Thank you for identifying this issue!

PR #641, opened 2020-06-22
Handle empty trailers case 

### LINKED ISSUE #638: hparser error in state on_trailers
Updating from hackney 1.15.2 to 1.16.0 http connections against a microhttpd-powered process randomly* fails with error (elixir syntax here)

```elixir
  {error: {:more,
   {:hparser, :response, 4096, 10, 0, :on_trailers, "", {1, 1}, "", [],
    :undefined, "chunked", "keep-alive", "application/json", "", :done}}}
```

Bisecting seems that the error appeared with commit https://github.com/benoitc/hackney/commit/78ce833a88715b195a045c6c949dca6ea1dc7d7a .

As additional info, if we force to *not* use keep-alive (by adding the header `Connection: close` on hackney) everything works ok.

Any idea on where to look?

`*` randomly means that we need to run several requests to get the error, sometimes 100 rqs are ok in a row, sometime we get 1 error in those 100, other times 6 errors and so on. Is reproducible but does not have a fixed pattern. Right now we're not been able to reproduce it against external services (google, our own apache/nginx servers, etc)

rozap:
* fixes #638 
* bug i think appeared in 6ddc33c1
* when streaming the response, we terminate
  the parse recursive calls early since
  there is no match for {:more, {hparser, ...:on_trailers...}}
  case

### DIFF
--- a/src/hackney_response.erl
+++ b/src/hackney_response.erl
@@ -147,6 +147,8 @@ stream_body(Client=#client{parser=Parser, clen=CLen, te=TE}) ->
 stream_body(Data, #client{parser=Parser}=Client) ->
   stream_body1(hackney_http:execute(Parser, Data), Client).
 
+stream_body1({more, Parser}, Client) ->
+  stream_body_recv(<<>>, Client#client{parser=Parser});
 stream_body1({more, Parser, Buffer}, Client) ->
   stream_body_recv(Buffer, Client#client{parser=Parser});
 stream_body1({ok, Data, Parser}, Client) ->


escobera:
This is affecting me as well.
Is there anything I can do to help merge this @benoitc ?

chess4ever (approved):

chess4ever:
@benoitc can we release this fix?

benoitc:
@chess4ever dependecies are released today. Hackney release will happen on Friday.

PR #653, opened 2020-08-21
add new url format for unix socket

benoitc:
Current way to pass the socket requires the user to url encode the path.
For example `/var/run/mysocket`should be encoded as
`%2Fvar%2Frun%2Fmyscoket` which is not really convenient.

Changes:

We introduce thiss tuple as a ne format `{<<"unix:/var/run/mysock'>>,<<"http://path/to/resource"> }`

This allows the user to pass a path withut encoding it while supporting
extensions in the future.

note: for now only http is suppored.

### DIFF
--- a/src/hackney.erl
+++ b/src/hackney.erl
@@ -345,7 +345,7 @@ request(Method, #hackney_url{}=URL0, Headers0, Body, Options0) ->
       Error
   end;
 request(Method, URL, Headers, Body, Options)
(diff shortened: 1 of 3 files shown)


benoitc:
Closing as outdated. The hackney codebase has been significantly refactored. The URL parsing logic has changed (uses `cut_fragment`/`cut_query` instead of `parse_fragment`) and `hackney.erl` request handling has been restructured. Thank you for the contribution!

Issue #654, opened 2020-08-25
Verify partial_chain with custom ssl_options


spydon:
If I'm not mistaken it is not possible to use the default `partial_chain` option when using custom ssl_options in 1.16.0 since `partial_chain/1` isn't exported anymore. 
Or is there a possibility to merge custom ssl options with the default ones?

benoitc:
`check_hostname_opts/1`is exported from hackney_ssl. Depending on what you need that maybe enough. I can provbably export the `partial_chain` callback  if needed. What require your custom ssl options for now?

spydon:
I think we mainly use a specific `verify_fun` and also swap `cacerts` to `cacertfile` in some cases.
The most useful would be if there was a way to use the hackney SSL defaults opts and then just override a few instead of all of them, is that possible somehow?

benoitc:
This has been resolved! Hackney now properly merges custom SSL options with the defaults.

When you provide `ssl_options`, hackney's `merge_ssl_opts/3` function merges your custom options with the secure defaults (including `partial_chain`, `verify_fun`, `cacerts`, etc.). Your custom options will override specific defaults while preserving the others.

Additionally, both `hackney_ssl:ssl_opts/2` and `hackney_ssl:check_hostname_opts/1` are exported, so you can:

```erlang
%% Get the default secure SSL options for a host
DefaultOpts = hackney_ssl:check_hostname_opts("example.com"),

%% Or just pass ssl_options and hackney merges for you
hackney:request(get, URL, [], <<>>, [
  {ssl_options, [{verify_fun, YourCustomFun}, {cacertfile, "/path/to/ca.pem"}]}
]).
```

The `cacertfile` case is also handled - if you provide `cacertfile`, hackney automatically removes `cacerts` to avoid conflicts.

Closing as resolved.

PR #655, opened 2020-08-26
Allow merging of SSL opts

spydon:
Related to #654
Since it isn't currently possible to merge your custom ssl opts with the default ones it could be a first step to expose the default ones so that the merging can be done client side.

### DIFF
--- a/src/hackney_connection.erl
+++ b/src/hackney_connection.erl
@@ -20,6 +20,7 @@
 
 -export([connect_options/3]).
 -export([ssl_opts/2]).
+-export([merge_ssl_opts/2]).
 
 -include("hackney.hrl").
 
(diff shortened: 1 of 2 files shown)


benoitc:
wouldn't exposing the `partial_chain` function just be enough?

spydon:
It would work, but then the client still needs to duplicate the full list of default otps when setting custom ones.

spydon:
What do you think @benoitc? 
And also, do you know why the checks are failing on some of the erlang versions?

benoitc:
> What do you think @benoitc?
> And also, do you know why the checks are failing on some of the erlang versions?

don't worry about the syntax issue. It's one of these usual  bugs of Travis...  The issue with that change is that some people will still wnt maybe to use some other settings imo. Maybe instead of replacing ssl options we could merge them or provide a way to merge defaults and the one in options ? this could be done there: https://github.com/benoitc/hackney/blob/master/src/hackney_connection.erl#L125 

Thoughts?

spydon:
Yeah, that would be a better solution. But that would either be a breaking change, or to introduce a new method there?
With the current change it will work the same as it does currently for those who want to use other settings, since the merging of the settings will need to be done client side and you would have to consciously call the new function to get the default opts.

spydon:
@benoitc would something like this be better maybe?

spydon:
Sorry to be a bother, have you had any time to look at this yet @benoitc? :)

benoitc:
sorry for the late answer. That would probably work. I am doing  a release today possibly including it :)

spydon:
@benoitc is the plan to get this out on the 14th of November like it says on the milestone date? :)

spydon:
@benoitc any plan for when the new release will be released?

benoitc:
@spydon next release will happen next week on thursday.  this feature should be part of it normally.

benoitc:
> @spydon next release will happen next week on thursday. this feature should be part of it normally.

that has been slightly delayed. It will happen this week, coming with a new documentation website.

spydon:
Any updates on the release? :)
We are seeing several teams having this problem now.

benoitc:
> Any updates on the release? :)
> We are seeing several teams having this problem now.

i will merge a minor release with it later today

PR #659, opened 2020-10-07
Improve metrics setup and docs

britto:
While building a custom metrics module for hackney I noticed a few inconsistencies in how some metrics are set up and documented. This PR addresses three of them. I've split them into separate commits to make it easier to review and eventually drop whatever may be irrelevant.

1. Fix the `queue_count` metric documentation and setup. It was being initialized as `queue_counter` in `metrics:new`;
2. Update pool metric docs to match their names. They are currently prefixed with `hackney_pool`, but show up prefixed with `hackney` in the docs;
3. Add documentation for 2 host-based pool metrics: `new_connection` and `reuse_connection`.

Update: Sorry, just noticed 1. is the same change proposed at #583, 2. and 3. are still valid though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -528,13 +528,13 @@ been started.
 
 #### Metrics per Pool
 
-|Name                          |Type       | Description                                                        |
-|------------------------------|-----------|--------------------------------------------------------------------|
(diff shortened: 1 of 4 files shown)


benoitc:
thanks!

PR #662, opened 2020-10-24
Prevent xref warning for unmaintained Erlang/OTP function

paulo-ferraz-oliveira:
I got an `xref` analysis error for an unknown function: `ssl:seed` is not documented in Erlang/OTP's doc.s as "maintained" (it's not there at all), so it's most surely been deprecated a good while ago. I leave it to your judgement to reject the PR (keeping current behavior with no more than a little itch to scratch) or accept it (thus embracing the present and the future).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ mime.types
 .rebar
 *.plt
 .rebar
+*_plt
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ before_install:
   - sudo pip install -q gunicorn
 
 script:
+  - ./support/rebar3 xref
   - ./support/rebar3 eunit
   - ./support/rebar3 dialyzer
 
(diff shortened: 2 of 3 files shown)


paulo-ferraz-oliveira:
Oh, I also took the liberty of adding `xref` to the CI steps, since now `hackney` is xref issue -free.

paulo-ferraz-oliveira:
`ping`

paulo-ferraz-oliveira:
`pang`

paulo-ferraz-oliveira:
Is this missing anything you can think of, @benoitc?

paulo-ferraz-oliveira:
Is this missing anything you can think of, @benoitc?

benoitc:
shouls be ok. I will take care of it in tomorrow release :)

benoitc:
thanks for the patch!

PR #666, opened 2020-11-25
Update netloc in client state when redirecting

mwcampbell:
This fixes a problem that caused `hackney:location/1` to return the wrong URL after a redirect if the redirect pointed to a different host.


benoitc:
thanks!

PR #669, opened 2020-12-08
Implement GitHub Actions

### LINKED ISSUE #668: move CI out of Travis-CI
This is not entirely related to latest changes in their terms but I would like to put the CI on another services to get faster and more consistent results .

paulo-ferraz-oliveira:
Main differences to current CI (Travis):
* using stock `rebar3` instead of "latest" (don't know if this is a hard requirement)

Closes #668.

### DIFF
--- a/.github/workflows/erlang.yml
+++ b/.github/workflows/erlang.yml
@@ -1,1 +1,35 @@
-
+---
+name: build
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  ci:
+    name: Run checks and tests over ${{matrix.otp_vsn}} and ${{matrix.os}}
+    runs-on: ${{matrix.os}}
+    container:
+      image: erlang:${{matrix.otp_vsn}}
+    strategy:
+      matrix:
+        otp_vsn: ["19.0", "19.3",
+                  "20.0", "20.1.7", "20.3.8.22",
+                  "21.0.9", "21.1.4", "21.2.7", "21.3.8.1",
+                  "22.0.7", "22.2.8", "22.3.4",
+                  "23.0.2", "23.2"]
+        os: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          apt-get -q update
+          apt-get -y install python3
+          apt-get -y install python3-pip
+          pip3 install httpbin
+          pip3 install gunicorn
+          gunicorn -b 127.0.0.1:8000 -b unix:httpbin.sock httpbin:app&
+      - run: ./support/rebar3 xref
+      - run: ./support/rebar3 eunit
+      - run: ./support/rebar3 dialyzer
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ __Version:__ 1.16.0
 
 **hackney** is an HTTP client library for Erlang.
 
-[![Build Status](https://travis-ci.org/benoitc/hackney.png?branch=master)](https://travis-ci.org/benoitc/hackney)
+[![Build Status](https://github.com/benoitc/hackney/workflows/build/badge.svg)](https://github.com/benoitc/hackney)
 [![Hex pm](http://img.shields.io/hexpm/v/hackney.svg?style=flat)](https://hex.pm/packages/hackney)
 
 ## Main features:


benoitc:
thanks! I will check later this morning. What should I do to enable github actions?

paulo-ferraz-oliveira:
> thanks! I will check later this morning. What should I do to enable github actions?

Two ways I can think of:
a. you merge to `master` - it'll be automatic, but this'll mean that whatever is not working will have to be fixed, either through `master` or a new pull request
b. you go to the Actions' tab, choose Erlang and commit an empty file to `master`; after that I'll rebase this pull request on top of `master` and we can make sure that what gets pushed was previously tested

Preferably you'd choose b) 😄

.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
...
+        os: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v2
+      - run: kerl list installations
+      - run: sudo pip install -q httpbin
+      - run: sudo pip install -q gunicorn
+      - run: "gunicorn -b 127.0.0.1:8000 -b unix:httpbin.sock httpbin:app&"
+      - run: rebar3 xref
+      - run: rebar3 eunit
+      - run: rebar3 dialyzer
```
aleDsz, line 26: You can use multline operations with you want to.

```suggestion
      - run: |
        kerl list installations
        sudo pip install -q httpbin
        sudo pip install -q gunicorn
        "gunicorn -b 127.0.0.1:8000 -b unix:httpbin.sock httpbin:app&"
        rebar3 xref
        rebar3 eunit
        rebar3 dialyzer
```

aleDsz (approved):
Just a suggestion ❤️

.github/workflows/ci.yml
paulo-ferraz-oliveira: Sure, but doesn't that make it a single task? I prefer separate tasks to easily identify where exceptions occur. Is there a "standard way of doing things"?

paulo-ferraz-oliveira: If the maintainers prefer it, though, I'm all for changing it.

paulo-ferraz-oliveira: We could also do `rebar3 do xref, eunit, dialyzer`, for example, but it's separated for the same reason as before. On the other hand, if it one day becomes possible to skip an action, if it's failing, visibility is increased even further.

aleDsz: Yeah, true. Just if maintainers prefer it <3

paulo-ferraz-oliveira: In any case, the suggestion is missing whitespace :)

benoitc: separate tasks makes maintenance easier imo :)

benoitc:
> > thanks! I will check later this morning. What should I do to enable github actions?
> 
> Two ways I can think of:
> a. you merge to `master` - it'll be automatic, but this'll mean that whatever is not working will have to be fixed, either through `master` or a new pull request
> b. you go to the Actions' tab, choose Erlang and commit an empty file to `master`; after that I'll rebase this pull request on top of `master` and we can make sure that what gets pushed was previously tested
> 
> Preferably you'd choose b) 😄

Sorry for the late answer. I will do a proper commit today so we can test it. Thanks !

benoitc:
> Preferably you'd choose b) 😄
done :) I pushed an empty action file.  Sorry for the delay... Go ahead anyway

paulo-ferraz-oliveira:
Rebased and force-pushed. 22.1.8.1 is not available. I'm removing it from the CI to check if the rest is OK.

paulo-ferraz-oliveira:
I'm removing `kerl list installations` from CI, since `kerl` is not available in the container. Is it absolutely required?

paulo-ferraz-oliveira:
@benoitc: let me know if you want some changes still. I'm using `rebar3` from the checkouts since some of the versions bundled with the containers have bugs that would otherwise impact CI for no good reason.

paulo-ferraz-oliveira:
Oh, you might also want to remove Travis CI from the app, since it's no longer serving any particular purpose.

benoitc:
> I'm removing `kerl list installations` from CI, since `kerl` is not available in the container. Is it absolutely required?

not at all :) I will merge the PR , thanks!

Issue #651, opened 2020-08-18
Unexpected hackney's messages in client process message queue


kpribylov:
Hi,

We've faced with the case when in some conditions hackney's messages might unintentionally appear in the client process message queue, e.g.:
{#Ref<0.2813583759.4118020097.25361>,{error,no_socket,<0.1006.0>}}    
  
Investigation has revealed that there is a kind of race condition when we try to get a socket from default hackney pool.  
Checkout request to 'hackney_pool' (sync call) is sent from client process and when it's timed-out (for any reason) the client has  '{error, checkout_timeout}' as return value and that's okay.
However, 'hackney_pool' might send a response for the previous request then and this message will be put directly to the client's message queue (see diagram below)

I have an idea to add a temporarily intermediary process between client app and hackney pool to handle the checkout procedure

![diagram-1](https://user-images.githubusercontent.com/50016497/90485774-a7f18000-e140-11ea-8dbd-56299a658413.png)

PR #670, opened 2020-12-09
Fix race condition within pool checkout procedure

kpribylov:
Hi,

Some time back I made a fix #652 for issue #651
The fix itself is fine, but it seems like we made a hurried decision to add a protective timeout for waiting checkout results https://github.com/benoitc/hackney/pull/652/commits/23d44cbbd42da2f4ddcd49466ece8cbff3285937

The problem is that there are two timers involved in the pool checkout procedure `ConnectTimeout` and `CheckoutTimeout`

Therefore we should wait at least `ConnectTimeout + CheckoutTimeout` to let the intermediary process to complete

However, even doing that we might end up sending unexpected ({checkout,#Ref<...>,{error,checkout_timeout}}) messages to the application process because of having more than one guard timer for the same purpose armed for the same timeout
To be absolutely sure we have to  `ConnectTimeout + CheckoutTimeout + SomeDelta` which is opaque

Having said that I suggest rolling back commit   https://github.com/benoitc/hackney/pull/652/commits/23d44cbbd42da2f4ddcd49466ece8cbff3285937 and keep managing all the timeouts inside do_checkout

Thanks

### DIFF
--- a/src/hackney_pool.erl
+++ b/src/hackney_pool.erl
@@ -59,17 +59,14 @@ start() ->
   ok.
 
 %% @doc fetch a socket from the pool
-checkout(Host, Port, Transport, #client{options = Opts} = Client) ->
-  ConnectTimeout = proplists:get_value(connect_timeout, Opts, 8000),
-  %% Fall back to using connect_timeout if checkout_timeout is not set
-  CheckoutTimeout = proplists:get_value(checkout_timeout, Opts, ConnectTimeout),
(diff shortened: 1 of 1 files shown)


kpribylov:
BTW, looks like that aliases will help us to solve the original problem gracefully
https://www.erlang.org/erlang-enhancement-proposals/eep-0053.html

kpribylov:
I've checked the fix on our load stress setup and it it seems that everything is fine...

Issue #665, opened 2020-11-24
:hackney_url.pathencode should use uppercase hex


smanolloff:
According to [RFC3986](https://tools.ietf.org/html/rfc3986):

> The uppercase hexadecimal digits 'A' through 'F' are equivalent to
   the lowercase digits 'a' through 'f', respectively.  If two URIs
   differ only in the case of hexadecimal digits used in percent-encoded
   octets, they are equivalent.  **For consistency, URI producers and
   normalizers should use uppercase hexadecimal digits for all percent-
   encodings.**

Since it is recommended to use upper-case, it's better to also do it here.

benoitc:
Good catch. 

That change is tricky to handle though at it my break some application relying on urls consistency or sort of by hashing them. The implementation will be done this way:

* add an option to path encoding to lowercase.
* default will be uppercase
* increment hackney minor release

benoitc:
implemented via #671 . I decided to just make it the default in urlencode and produce uppercase using pathencode. The change will be documented as a braking change in the changelog.

benoitc:
fixed in master. Thanks for the ticket!

PR #671, opened 2020-12-16
breaking change: produce uppercase hexadecimal in urls

benoitc:
This change the behaviour of urlencode and pathencode to produce
uppercase hexadecimal following the RFC3986:

    The uppercase hexadecimal digits 'A' through 'F' are equivalent to
    the lowercase digits 'a' through 'f', respectively. If two URIs
    differ only in the case of hexadecimal digits used in percent-encoded
    octets, they are equivalent. For consistency, URI producers and
    normalizers should use uppercase hexadecimal digits for all percent-
    encodings.

breaking change that requires to bump the minor release.

### DIFF
--- a/src/hackney.app.src
+++ b/src/hackney.app.src
@@ -4,7 +4,7 @@
 {application, hackney,
   [
     {description, "simple HTTP client"},
-    {vsn, "1.16.0"},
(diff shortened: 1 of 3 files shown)


benoitc:
fix #665

PR #686, opened 2021-05-19
make checkout synchronous again

benoitc:
This changes remove the intermediate process to checkout a process. When
we reply asynchronously we also check if the proces that will receive
the message is alive.. This should remove any unwanted messages when a
request timeout.

Changes:

 * remove intermediate process in hackney_pool:checkout/4
 * send async reply only if the requester is alive


benoitc:
fix #683

PR #691, opened 2021-09-03
Fixed slow unix domain socket requests

dparnell:
Hi,

I have been using HTTPoison and by extension hackney in a project and found that when making requests to a Unix domain socket sometimes it would be excruciatingly slow. I finally had some time to track it down and discovered that hackney was doing a DNS lookup on the domain socket path which of course would always fail, but sometimes take a long time to do so.
This little patch fixes this problem.

Thankyou so much for making hackney :)

Daniel


benoitc:
thanks for the patch. merged.

PR #692, opened 2021-10-11
Follow HTTP 308 redirects

### LINKED ISSUE #635: Follow HTTP 308 redirects
Hi,

Hackney specifies the following status codes as redirects follow with follow_redirect: https://github.com/benoitc/hackney/blob/6bd1f46d98757246cd0896624ff12587451ed122/src/hackney.erl#L735

I think [HTTP 308](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308) should be included too. For instance, pinterest returns HTTP 308 when accessing the non-www domain.

AntoineAugusti:
Fixes #635 

I'm not sure if I need to make other changes. Did a few searches on the codebase and did not find obvious things to change. Let me know if I missed something.

### DIFF
--- a/src/hackney.erl
+++ b/src/hackney.erl
@@ -725,7 +725,7 @@ maybe_redirect(
   case redirect_location(Headers) of
     undefined -> Resp;
     Location ->
-      IsRedirect = lists:member(S, [301, 302, 303, 307]),
+      IsRedirect = lists:member(S, [301, 302, 303, 307, 308]),
       case IsRedirect of
         false -> Resp;
         _ ->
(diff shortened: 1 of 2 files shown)


AntoineAugusti:
@benoitc Wondering if we can do this regarding this PR!

AntoineAugusti:
Hi @benoitc, we would love to get this change merge as we require it upstream ❤️

fchabouis:
Hi,
I would be glad too if this was merged!
Thanks :)

benoitc:
there is a coming release tomorrow. This will be part of it.

Issue #695, opened 2021-10-29
Hackney should ignore 1XX responses


derek-zhou:
Some website send 103 response before the 200, even in http1.1. One such is http://www.theregister.com
It could be a cloudflare thing. 

hackney 1.8.0:

```
iex(3)> :hackney.request("https://www.theregister.com/")

{:ok, 103,
 [
   {"Link",
    "<https://securepubads.g.doubleclick.net/tag/js/gpt.js>; as=\"script\"; rel=\"preload\", </design_picker/aec273bc80dd0dc3a73edce687f7cdaa0e9ef0f5/javascript/_.js>; as=\"script\"; rel=\"preload\", </css_picker/webkit/aec273bc80dd0dc3a73edce687f7cdaa0e9ef0f5/scaffolding.css>; as=\"style\"; rel=\"preload\", </css_picker/webkit/aec273bc80dd0dc3a73edce687f7cdaa0e9ef0f5/design.css>; as=\"style\"; rel=\"preload\", </design_picker/5e49edbd1875f214e0decae1e24b200066780fa8/style/fonts/arimo/arimo-700.latin.woff2>; as=\"font\"; crossorigin; rel=\"preload\", </design_picker/5e49edbd1875f214e0decae1e24b200066780fa8/style/fonts/arimo/arimo-400.latin.woff2>; as=\"font\"; crossorigin; rel=\"preload\""}
 ], #Reference<0.2513840446.3424649217.237339>}
```

If I edit the hackney source and force HTTP/1.0, the server will only response with one 200 response and everything will turn out ok. 

I think the correct change for hackney is to throw away all 1XX response because there is no way to expose them to the caller anyway. The subsequent 2XX response is the meat the caller want.

leonardb:
Of interest for this: https://blog.cloudflare.com/early-hints/

Looks like these will probably become more prevalent as time goes by and should be supported asap to prevent breakage

PR #696, opened 2021-10-29
add junk line skipping for 1XX responses

derek-zhou:
for #695 
Any response code < 200 shall be ignored, together with whatever header lines there are.


derek-zhou:
Ping. BTW: I've been running with this patch in production for a week.

benoitc:
sorry for the delay to apply. I will make an  versio nof hackney in february with more deeper changes also.

Issue #698, opened 2021-11-19
Connection timeout errors


SukhikhN:
Hello!

I'm testing my application behaviour with different network timeouts and getting strange error results when unsig hackney with or without connection pool.

First, for testing connection timeout I set up iptables to drop all packets incoming to port 8080:
`# iptables -I INPUT -p tcp --dport 8080 -j DROP`

Then I try to make requests from the application console:
```
iex(2)> :hackney.request(:get, "http://localhost:8080", [], "", [pool: false])
{:error, :connect_timeout}
iex(3)> :hackney.request(:get, "http://localhost:8080", [], "", [])           
{:error, :timeout}
```
As you can see the error reasons are different depending on pool configuration, which seems strange to me - is this correct behaviour?

Also `:timeout` reason is returned when server is slow to respond - so we can't distinguish response timeout from connection timeout.

I'm testing it using hackney 1.18.0.

benoitc:
closing as fixed. sorry for the late handling.

PR #700, opened 2021-12-20
Restore connect timeout error

sa-mm:
This error seems to have disappeared with updates in 1.16.0. This change makes errors related to connection timeouts consistent whether one is using a pool or not. User reported issue here: [benoitc#698](https://github.com/benoitc/hackney/issues/698)


benoitc:
thanks!

PR #703, opened 2022-03-16
Correctly handle badly formatted protocols

AntoineDPigeon:
Remove the bad protocol from the URL instead of parsing that protocol inside the domain.


benoitc:
Late response, but what are you trying to fix?

PR #706, opened 2022-06-28
(feat) Support LF separators since rfc7230-3.5 allows for LF-only

leonardb:
```Although the line terminator for the start-line and header fields is
   the sequence CRLF, a recipient MAY recognize a single LF as a line
   terminator and ignore any preceding CR.```

I'm running into this issue with an SMS service provider's API where all responses use LF instead of CRLF in the headers.


benoitc:
merged . Thanks for the change

PR #709, opened 2022-09-21
Use correct SPDX ID for licenses field

Frost:
The "Apache License 2.0" has the SPDX ID "Apache-2.0", according to the list at https://spdx.org/licenses/. While the string "Apache 2.0" is fully readable for humans, some tools use these fields to verify what licenses are used, and the tools tend to not be as flexible when reading strings that are interpreted as identifiers.


benoitc:
thanks!

PR #734, opened 2024-02-13
HTTP and HTTPS proxies might differ

Juliusan:
Hackney handle `HTTP_PROXY` and `HTTPS_PROXY` variables as they define the same thing. In fact (e.g. `curl` documentation) `HTTP_PROXY` should be used only for http requests and `HTTPS_PROXY` should be used only for https requests. This PR fixes that.
Moreover, this PR adds `ALL_PROXY` environment variable handling.

### DIFF
--- a/include/hackney.hrl
+++ b/include/hackney.hrl
@@ -57,5 +57,5 @@
 
 -define(CONNECTIONS, hackney_connections).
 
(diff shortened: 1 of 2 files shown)


benoitc:
thanks!

Issue #733, opened 2024-02-05
`secure_renegotiate` and `reuse_sessions` is incompatible with `tlsv1.3`


Juliusan:
While trying to connect to web server using TLS v1.3 using Hackney, I get following errors:
```
{options,incompatible,[secure_renegotiate,{versions,['tlsv1.3']}]}
{options,incompatible,[reuse_sessions,{versions,['tlsv1.3']}]}
```
Apparently, Hackney adds default values for `secure_renegotiate` and `reuse_sessions` parameters (https://github.com/benoitc/hackney/blob/f190daf67cfefc4f1950011759624f8f87260597/src/hackney_ssl.erl#L139-L141) while connecting via ssl. However Erlang's ssl code fails if these parameters are mentioned regardless of their values. As Hackney user you can override the value of these parameters, however it is impossible to override passing them to Erlang's ssl library.

I am using Erlang/OTP 26.2, Hackney version 1.20.1 (commit `f190daf67cfefc4f1950011759624f8f87260597`). My additional parameters passed to `hackney:post` are:
```
[
    {connect_timeout, 50000},
    {recv_timeout,    50000},
    with_body,
    {ssl_options,[{versions, ['tlsv1.3']},{verify,verify_none}]}
]
```

Is there a way to avoid passing these parameters to `ssl:connect`?

benoitc:
my bad , I didn't see your ticket. I just removed them from the base options in the sake of simplicity. thanks for the patch anyway.

PR #735, opened 2024-02-13
Fix options that conflict with OTP's ssl library

Juliusan:
This is fix for issue issue #733. More details are provided there.


benoitc:
my bad , I didn't see your ticket.  I just removed them from the base options in the sake of simplicity. thanks for the patch anyway.

PR #737, opened 2024-02-22
improve connection to remote

benoitc:
This pR track different changes to provide a better UX during connection to remote.

- [x] happy-eyeball support
- [x] full support of proxy configuration from the system environnement


src/hackney_happy.erl
```diff
@@ -0,0 +1,72 @@
...
+    false ->
+      case hackney_cidr:is_ipv4(Hostname) of
+        true ->
+          ?report_debug("connect using IPv4", [{hostname, Hostname}, {port, Port}]),
+          gen_tcp:connect(Hostname, Port, [inet | Opts], Timeout);
+        false ->
+          ?report_debug("happy eyeballs, try to connect using IPv6",  [{hostname, Hostname}, {port, Port}]),
+          Self = self(),
+          Addrs = getaddrs(Hostname),
+          Pid = spawn_link( fun() -> try_connect(Addrs, Port, Opts, Self, {error, nxdomain}) end),
```
benoitc, line 28: no need to spawn here, ,,either wya makes DNS resolution asynchronous.

benoitc:
I should provide a way to bypass DNS resolution and allows the user to force the way it wants to make a connection. Also allow them to choose the preference (for those that have to support the legacy with IPv4).

src/hackney_happy.erl
```diff
@@ -0,0 +1,77 @@
...
+                  erlang:demonitor(MRef, [flush]),
+                  {error, connect_timeout}
+          end
+      end
+  end.
+ 
+getaddrs(Hostname) ->
+  IP6Addrs = [{Addr, 'inet6'} || Addr <- getbyname(Hostname, 'aaaa')],
+  IP4Addrs = [{Addr, 'inet'} || Addr <- getbyname(Hostname, 'a')],
+  IP6Addrs ++ IP4Addrs.
```
ruslandoga, line 46: 👋 @benoitc 

I wonder if it would make sense to "intersperse" the addresses here like [ipv6, ipv4, ipv6, ipv4, etc.]?

ruslandoga: Sort of like this: https://github.com/ruslandoga/happy_tcp/blob/afbf889d566e6de743f66978f44b831987acbd6e/lib/happy_tcp.ex#L61-L81

PR #746, opened 2024-09-06
Fix intermittent chunked response hang

danielfinke:
When streaming a chunked response, it is possible to cause a TCP receive hang under particular circumstances. If at one point the parser buffer doesn't have the whole chunk, at a later point the buffer ends up empty `<<>>`, and subsequently `hackney_response:stream_body/1` is called, `hackney_response:recv/2` will hang if the expected remaining size exceeds the remainder of the response. That expected size is actually stale, from the earlier point when the parser did not have the whole chunk. This issue slipped in with benoitc/hackney#710.

This was identified when using https://github.com/benoitc/couchbeam and sending several chunked requests. If the last non-terminating chunk completed the response JSON object, `hackney_response:skip_body/1` is called to discard the remaining body, but is told to receive a number of bytes equal to the expected remaining size which will frequently exceed the small terminating chunk and trailers. As a result, the recv operation hangs waiting for bytes that will never arrive.

Now, the transfer state (`BufSize`/`ExpectedSize`) are reset after each successful chunk. The speed benefit of benoitc/hackney#710 is retained (tested with the same approach as in that PR).

- correct some related typespecs

### DIFF
--- a/src/hackney_http.erl
+++ b/src/hackney_http.erl
@@ -400,33 +400,27 @@ parse_body(St) ->
 
 
 -spec transfer_decode(binary(), #hparser{})
-    -> {ok, binary(), #hparser{}} | {done, binary()} | {error, atom()}.
+    -> body_result() | {more, binary()} | {error, atom()}.
 transfer_decode(Data, St=#hparser{
   body_state={stream, TransferDecode,
-    TransferState, ContentDecode},
+    TransferState, ContentDecode}=BodyState,
   buffer=Buf}) ->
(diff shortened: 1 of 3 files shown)


jamesaimonetti:
@benoitc this has been working well for us lately. Curious if you have a chance to review?

benoitc:
on it sorry for the delay.

danielfinke:
@benoitc hey there, anything needed from our end? I double-checked and I see the same EUnit failures on `master` as on this branch, so I don't believe they are a result of my changes.

PR #750, opened 2025-01-22
Treat iolist of empty binaries as empty body

pvsr:
Normally when the request body is empty, `handle_body` is only called for POST and PUT requests, since other methods shouldn't have `Content-Type` and `Content-Length` headers for empty bodies. But one case is missed in the empty body detection, an list containing empty binaries. The [gleam hackney wrapper](https://github.com/gleam-lang/hackney) represents empty bodies as `[<<>>]`, triggering this bug and causing https://github.com/gleam-lang/hackney/issues/5.

This PR expands the empty body detection to check if the result of `iolist_to_binary` is empty.


src/hackney_request.erl
```diff
@@ -348,13 +353,6 @@ handle_body(Headers, ReqType0, Body0, Client) ->
                             S = hackney_headers_new:get_value(<<"content-length">>, Headers),
                             {S, CT, Body0};
 
-                          _ when is_list(Body0) -> % iolist case
```
pvsr, line 351: This branch isn't needed anymore since an iolist body will now be converted a to binary before being passed to `handle_body`.

src/hackney_request.erl
```diff
@@ -348,13 +354,6 @@ handle_body(Headers, ReqType0, Body0, Client) ->
...
```
pvsr, line 351: This case is now handled before calling `handle_body`.

pvsr:
I rebased and added some integration tests.

PR #776, opened 2025-07-23
Respect connect timeout in ssl:connect

savonarola:
After socket connection called with timeout, `ssl:connect` (the handshake) is called without timeout. But the handshake may last long and easily exceed the expected connect timeout.  So we use `ssl:connect/3` to limit the handshake time too.


benoitc:
Thank you. I've updated your PR to take care of the infinity case in #781 . However i'm wondering if i's acually makes sense to do this since if we have now a socket why would we timeout ?  Do you have any specific use case in mind?

PR #788, opened 2025-11-07
Fix socket unused timeouts taking 100ms

nigel-rook-happening:
`hackney_pool` currently blocks for 100ms every time it closes a connection due to it being unused for the configured timeout.

The flow is as follows:

1. hackney gets a socket checkin, it creates a timer which triggers if nothing uses the socket for 150s (default)
2. timer expires, without the socket being used again, hackney_pool gets a `{timeout, Socket}` message
3. `handle_info({timeout, Socket}, ...)` calls `remove_socket(Socket, ...)`, which calls `cancel_timer` for the associated timer
4. in `cancel_timer`, `erlang:cancel_timer` calls false, since the timer has already expired. `cancel_timer` then waits for the `{timeout, Socket}` message or 100ms. hackney_pool is handling that message already, so it's not gonna arrive, so we wait 100ms

We've got a pretty high request rate, so when hackney_pool goes unresponsive for 100ms, we get a surge of backed up checkout requests. That results in more connections being created because there aren't enough in the pool to deal with the backed up queue. That in turn leads to more timeouts after 150s. So the whole thing cascades out of control until we get lots of `checkout_timeout` errors.

This PR adds an extra param to `remove_socket` to tell it to not cancel the timer. That param is used when receiving the `{timeout, Socket}` message, in order to prevent any attempt to cancel the already expired timer.

### DIFF
--- a/src/hackney_pool.erl
+++ b/src/hackney_pool.erl
@@ -340,7 +340,7 @@ handle_cast(_Msg, State) ->
   {noreply, State}.
 
 handle_info({timeout, Socket}, State) ->
-  {noreply, remove_socket(Socket, State)};
+  {noreply, remove_socket(Socket, State, false)};
 handle_info({tcp, Socket, _}, State) ->
   {noreply, remove_socket(Socket, State)};
 handle_info({tcp_closed, Socket}, State) ->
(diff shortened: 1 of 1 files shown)


nigel-rook-happening:
And, now is when github tells me that you already have a fix for this in master. I'll close this for you :)

PR #789, opened 2025-11-13
remove sort of dns responses

albertasi:
This PR stops sorting the resolver’s address list in hackney_happy:getbyname/2 by replacing lists:usort(AddrList) with AddrList. The current sorting removes duplicates and reorders IPs, which unintentionally breaks DNS randomness and can cause uneven connection distribution.

Rationale
DNS randomness is a common, simple load-spreading mechanism. Reordering results undermines it.
Duplicate handling is typically the resolver’s responsibility. If duplicates occur, they are rare and harmless; removing them here is not worth the cost of losing order.


benoitc:
Closing as already fixed. The current master no longer uses `lists:usort(AddrList)` - the code now returns `AddrList` directly, preserving DNS order. Thank you for identifying this issue!

PR #829, opened 2026-02-26
fix: Invalid match in handle_h2_frame/2

### LINKED ISSUE #828: [Bug] Invalid match in handle_h2_frame/2 for bodies larger tham 65k bytes
When the HTTP/2 machine has queued DATA frames waiting for flow control window space, processing a WINDOW_UPDATE frame returns {send, SendList, H2Machine} to flush the queued data. 

Both the connection-level and stream-level WINDOW_UPDATE handlers in handle_h2_frame/2 only matched
{ok, ...} and {error, ...}, causing a case_clause crash for any HTTP/2 request body larger than the initial flow control window.

ycastorium:
Fixes #828


benoitc:
Thanks!

Issue #830, opened 2026-02-27
[Bug]: Off-by-one in HPACK dynamic table index calculation breaks HTTP/2 connection reuse


ycastorium:
The HPACK encoder in `hackney_hpack.erl` has an off-by-one error when converting absolute dynamic table indices to HPACK relative indices. This causes HTTP/2 requests to fail on any connection that reuses an existing HTTP/2 connection with a populated HPACK dynamic table.

The first request on a new connection works fine. The second request fails because the encoder emits incorrect indexed references that point to static table entries instead of dynamic table entries, causing pseudo-headers (`:method`, `:path`, etc.) to appear after regular headers in the encoded block.

How to reproduce:

```erlang
%% Connect to any HTTP/2 server
{ok, Ref} = hackney:connect(hackney_ssl, "httpbin.org", 443, [{pool, false}, {protocols, [http2]}]),

%% First request succeeds (headers encoded as literals)
{ok, 200, _, _} = hackney_conn:request(Ref, <<"GET">>, <<"/">>, [{<<"host">>, <<"httpbin.org">>}], <<>>),

%% Second request fails -- server rejects pseudo-header ordering
{error, _} = hackney_conn:request(Ref, <<"GET">>, <<"/">>, [{<<"host">>, <<"httpbin.org">>}], <<>>).
```

PR #831, opened 2026-02-27
fix: Fixes off-by-one error in HPACK

ycastorium:
Fixes: #830


benoitc (approved):
Thanks for the fix!  Good catch and nice test coverage :)

PR #838, opened 2026-04-05
fix: no_proxy parsing bugs

rda1902:
Fix: no_proxy environment variable parsing bugs
This PR fixes issues with parsing the no_proxy environment variable, based on version 1.25.0.
Problem:
When NO_PROXY contains multiple comma-separated hosts, only the last entry is honored. All preceding entries are silently ignored. 

Note for maintainer:
This is branched from 1.25.0 as i am unable to upgrade to the latest verstion at this time.

### DIFF
--- a/src/hackney.erl
+++ b/src/hackney.erl
@@ -26,7 +26,9 @@
          cancel_request/1,
          setopts/2]).
 
--export([redirect_location/1]).
+-export([redirect_location/1,
+         parse_no_proxy_env/2,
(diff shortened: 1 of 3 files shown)


benoitc:
Hi, Thank you for the patch. This is normally fixed in last. version of Hackney: https://github.com/benoitc/hackney/blob/master/src/hackney.erl#L1391

Any reason to not upgrade ?

benoitc:
fixed in latest version

PR #859, opened 2026-05-31
Fix dialyzer specs

tomekowal:
Hey! Sorry for a PR without corresponding issue, but I think it is a small change and non controversial change.

During work on Tesla support, I've seen dialyzer complaining. In a couple of places there were leftover `reference` types instead of currently used `pid` types. This PR fixes that.


benoitc:
merged. Thanks!

Issue #893, opened 2026-06-29
Unable to use Hackney 4.x with opentelemetry_exporter


seantanly:
The `opentelemetry_exporter` has a transitive dependency on chatterbox, whose module names `h2_*` clash with h2. 

```
opentelemetry_exporter <- grpcbox <- chatterbox
```
Unfortunately, both ' chatterbox` and `grpcbox` do not appear to be actively maintained. Will it be possible to rename the newly introduced h2 library such that the module naming conflict situation can be resolved swiftly?

Related: https://github.com/joedevivo/chatterbox/issues/154

Very much appreciated for looking into this. This will allow people who depend on both to upgrade their hackney version to resolve the recent CVEs.

walki:
We are running into this exact issue this morning with resolving recent CVEs.

benoitc:
Thank you for the ticket. You have 2 possibilities today to fix th eissue:

review the PR on grpcbox https://github.com/tsloughter/grpcbox/pull/120 replacing chatterbox by erlang_h2. This is the fastest path as the change will be transparent for all apps using chatterbox. Also erlang_h2 is actively maintained .

move to [livery_grpc](https://github.com/benoitc/livery_grpc). Which is fully supported GRPC specs and based on the livery framework.
Thanks for the patch and the head up anyway, that was very useful.

closing the issue.

PR #894, opened 2026-06-29
Use eh2_* namespace from renamed erlang_h2 (companion to erlang_h2#19, refs #893)

walki:
## Use the `eh2_*` namespace from renamed erlang_h2

Companion to **benoitc/erlang_h2#19**; part of fixing #893.

### Why
hackney 4.x bundles `erlang_h2`, whose `h2_*` module names collide with `chatterbox` (pulled via `grpcbox` ← `opentelemetry_exporter`). Because BEAM module names are global atoms, any app depending on both hackney 4.x and `opentelemetry_exporter` can't assemble a release:

```
** (Mix) Duplicated modules:
    h2_settings / h2_frame / h2_connection / h2_client specified in chatterbox and h2
```

erlang_h2#19 renames those modules to the `eh2_*` namespace. This is the matching hackney-side update.

### Change
hackney references erlang_h2 only through the connection module, so this is small:
- `h2_connection` → `eh2_connection` (27 call sites in `hackney_conn.erl` and `hackney_h2_stream.erl`).

The public `h2` API module, the `{h2, Conn, ...}` message tag, the OTP app name `h2`, and the dialyzer `plt_extra_apps` entry are all unchanged.

### Dependency note
This branch temporarily points the `h2` dep at the rename branch so it builds and can be verified end-to-end:

```erlang
{h2, {git, "https://github.com/walki/erlang_h2.git", {branch, "fix-h2-module-namespace-collision"}}}
```

Once erlang_h2#19 is merged and released, this should go back to the hex version (`{h2, "~> <renamed release>"}`). Happy to update once you cut that release.

### Verification
Built end-to-end against a real downstream app (Phoenix + `opentelemetry_exporter` + ex_aws on hackney): `rebar3 compile` clean, and the consuming app's `mix release` now assembles with **no duplicated-modules error**.

`webtransport` also references the renamed modules (`h2_varint`, `h2_capsule`) and would need the same one-line-namespace update; flagged in erlang_h2#19. Glad to send that PR too.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -55,7 +55,7 @@
         %% Pure Erlang QUIC + HTTP/3 stack
         {quic, "~>1.6.5"},
         %% Pure Erlang HTTP/2 stack
-        {h2, "~>0.10.1"},
+        {h2, {git, "https://github.com/walki/erlang_h2.git", {branch, "fix-h2-module-namespace-collision"}}},
         %% WebTransport client (HTTP/3 and HTTP/2) - powers the wt_* API
         {webtransport, "~>0.4.1"},
         {idna, "~>7.1.0"},
--- a/src/hackney_conn.erl
+++ b/src/hackney_conn.erl
@@ -195,7 +195,7 @@
     protocol = http1 :: http1 | http2 | http3,
     %% HTTP/2 connection pid (from h2 library)
     h2_conn :: pid() | undefined,
-    %% Monitor ref for the h2_connection process so hackney_conn fails fast
+    %% Monitor ref for the eh2_connection process so hackney_conn fails fast
     %% if the h2 lib crashes.
     h2_mon :: reference() | undefined,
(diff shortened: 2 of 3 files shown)


benoitc:
Hrm at this point it may be hard to change it . Why not migrating to livery_grpc instead ? https://github.com/benoitc/livery_grpc

benoitc:
I will close this PR since it's too late to rename H2. It's used in many libraries already and some of internal code here rely on it. 

You have 2 possibilities today to fix th eissue:

1. review the PR on grpcbox  https://github.com/tsloughter/grpcbox/pull/120 replacing chatterbox by erlang_h2. This is the fastest path as the change will be transparent for all apps using chatterbox. Also erlang_h2 is actively maintained .
2. move to [livery_grpc](https://github.com/benoitc/livery_grpc). Which is fully supported GRPC specs and based on the livery framework.

Thanks for the patch and the head up anyway, that was very useful.

PR #903, opened 2026-07-10
Stop non-reusable hackney_conn processes when a sync request completes (#902)

### LINKED ISSUE #902: hackney_conn processes leak: completed no_reuse (Connection: close) HTTP/1.1 connections park in `connected` forever (sync body-completion path never closes non-reusable connections)
> **Update (root cause corrected).** My first hypothesis below — that the leak is caused by `?IDLE_TIMEOUT` defaulting to `infinity` — turned out to be **incomplete**. Giving `?IDLE_TIMEOUT` a finite value does **not** fix the leak: the idle reaper does `{next_state, closed, Data}`, but `closed(enter)` for a **non-pooled** connection (`pool_pid = undefined`) simply `{keep_state, ...}` with no `stop` and no grace timer (only *pooled* connections arm `closed_grace_expired -> {stop, normal}`). So a finite timeout just moves the zombie from `connected` to `closed` — the socket closes, but the gen_statem process is never terminated and keeps accumulating.
>
> **The actual root cause is in the synchronous body-completion path.** `receiving({call, From}, body, ...)` and `receiving({call, From}, stream_body, ...)` unconditionally return `{next_state, connected, ...}` after the body is read — they never consult `no_reuse` or `should_close_connection/1`. So every completed `no_reuse` (or `Connection: close`) connection parks back in `connected` forever, pinned alive by its owner link → one leaked `hackney_conn` process per request. The **async** streaming path already handles this correctly in `finish_async_streaming/1` (it stops non-reusable / non-pooled connections); the **sync** path (what HTTPoison uses) does not. See the corrected "Root cause" and "Suggested fix" sections below.

### Summary

Under sustained load against an HTTP/1.1 endpoint whose connections are flagged `no_reuse` (e.g. `Connection: close`), `hackney_conn` gen_statem processes accumulate without bound and are **never terminated**. Each completed connection parks in the `connected` state as a child of `hackney_conn_sup` and nothing ever reaps it. Process memory grows until the node OOMs.

Observed on **4.4.5**; the relevant code paths are unchanged in **4.5.2**, so this is not fixed by upgrading.

### Live evidence (production node — plain-HTTP JSON-RPC client, a few req/s)

- `2539` `hackney_conn` processes and climbing ~1.5/s (was `1714` a few minutes earlier), while only **34** OS ports exist. These are orphaned FSMs, not live connections (so it is *not* an fd/socket leak).
- Every one of them:
  - `:sys.get_state/2` → `connected`
  - `hackney_conn:checkin_info/1` →
    ```erlang
    #{protocol => http1, upgraded_ssl => false, no_reuse => true,
      ready => false, should_close => false, pool_ssl => false}
    ```
- Their only live link is `hackney_conn_sup`. The calling owner is a long-lived process that never dies, so no owner-`DOWN` ever triggers cleanup.

### Root cause (in `src/hackney_conn.erl`)

The **synchronous** request path never closes a non-reusable connection when the request completes. After the body is fully read:

```erlang
receiving({call, From}, body, Data) ->
    case read_full_body(Data, <<>>) of
        ...
        {ok, Body, NewData} ->
            %% Socket still valid - return to connected state
            {next_state, connected, NewData, [{reply, From, {ok, Body}}]};   %% <-- unconditional
        ...
```

`stream_body` does the same on `{done, NewData}`. Neither consults `no_reuse` nor `should_close_connection/1`, so a `no_reuse`/`Connection: close` connection returns to `connected` and stays there. Because its owner is long-lived and its socket is idle (or already gone with no `{tcp_closed, _}` delivered), there is **no path to termination**.

Contrast the **async** path, which gets it right — `finish_async_streaming/1`:

```erlang
case should_close_connection(Data) of
    true -> ... {stop, normal, ...};
    false when PoolPid =:= undefined -> ... {stop, normal, ...};
    false -> {next_state, connected, reset_async(Data)}
end
```

Two secondary factors make it impossible for any fallback to save the parked process:

1. `-define(IDLE_TIMEOUT, infinity).` and `idle_timeout` is never populated from app-env or a request option, so the `connected(state_timeout, idle_timeout) -> {next_state, closed}` reaper is never armed.
2. Even if it were, `closed(enter, ...)` for a non-pooled connection (`pool_pid = undefined`) does `{keep_state, ...}` — it closes the socket but never `stop`s the process. Only pooled connections arm `closed_grace_expired -> {stop, normal}`.

This looks adjacent to #886 (HTTP/1.1 active-once stranding) and #888 (keepalive pooling correctness), but those addressed *reused/pooled* connections; the *completed non-reused* sync case is still uncovered.

### Reproduction

Issue `hackney:request/5` (e.g. via httpoison) repeatedly against a plain-HTTP/1.1 server that responds with `Connection: close`, at a modest rate, with a long-lived caller. Then count connection processes:

```erlang
length([P || P <- erlang:processes(),
             {dictionary, D} <- [erlang:process_info(P, dictionary)],
             match =:= (case lists:keyfind('$initial_call', 1, D) of
                          {_, {hackney_conn, _, _}} -> match; _ -> nomatch end)]).
```

It grows monotonically while open sockets (`length(erlang:ports())`) stay flat.

### Suggested fix

Make the sync completion path mirror `finish_async_streaming/1` — stop the process instead of parking it when the connection must not be reused:

```erlang
finish_sync_request(From, Reply, #conn_data{transport = Transport, socket = Socket,
                                            no_reuse = NoReuse} = Data) ->
    case NoReuse orelse should_close_connection(Data) of
        true ->
            case Socket of
                undefined -> ok;
                _ -> Transport:close(Socket)
            end,
            {stop_and_reply, normal, [{reply, From, Reply}],
             Data#conn_data{socket = undefined}};
        false ->
            {next_state, connected, Data, [{reply, From, Reply}]}
    end.
```

…called from both `receiving({call,From}, body, ...)` `{ok, Body, NewData}` and `receiving({call,From}, stream_body, ...)` `{done, NewData}`. Reusable keep-alive connections still return to `connected`.

Optionally, as defense-in-depth, also make `closed(enter, ...)` for `pool_pid = undefined` arm a grace timer + `{stop, normal}` (mirroring the pooled path) so that connections reaching `closed` via server-close / `tcp_error` also terminate.

Happy to send a PR.

lucianoengel:
Fixes #902.

### Problem

The **synchronous** body-completion path never closes a non-reusable connection when the request completes. After the body is fully read, `receiving({call, From}, body, ...)` and `receiving({call, From}, stream_body, ...)` both return `{next_state, connected, ...}` **unconditionally** — they never consult `no_reuse` or `should_close_connection/1`:

```erlang
{ok, Body, NewData} ->
    %% Socket still valid - return to connected state
    {next_state, connected, NewData, [{reply, From, {ok, Body}}]};
```

So a connection flagged `no_reuse` (proxy tunnels, SSL upgrades, pool disabled) or carrying `Connection: close` returns to `connected` and stays there. When the calling owner is long-lived (a pool worker, a persistent client process), no owner-`DOWN` ever fires, the idle socket produces no `{tcp_closed, _}`, and there is **no path to termination** — one `hackney_conn` gen_statem process leaks per request until the node OOMs.

The **async** path already handles this correctly in `finish_async_streaming/1` (it stops non-reusable / non-pooled connections). This PR brings the sync path to parity.

### Fix

Add `finish_sync_request/3`, mirroring the async logic, and call it from both the `body` and `stream_body` completion clauses:

```erlang
finish_sync_request(From, Reply, #conn_data{transport = Transport, socket = Socket,
                                            no_reuse = NoReuse} = Data) ->
    case NoReuse orelse should_close_connection(Data) of
        true ->
            case Socket of
                undefined -> ok;
                _ -> Transport:close(Socket)
            end,
            {stop_and_reply, normal, [{reply, From, Reply}],
             Data#conn_data{socket = undefined}};
        false ->
            {next_state, connected, Data, [{reply, From, Reply}]}
    end.
```

Reusable keep-alive connections still return to `connected`.

Note on the discriminator: I key on `no_reuse orelse should_close_connection/1` rather than copying `finish_async_streaming/1`'s `should_close orelse pool_pid == undefined`. In practice the leaking connections were `no_reuse: true` **and** pooled (`pool_pid` set, `should_close: false`), so the async condition would not have caught them — `no_reuse` is the correct signal for "this connection must not be reused." (Arguably `finish_async_streaming/1` should check `no_reuse` too, but I've kept this PR scoped to the observed leak.)

### Evidence

Observed on a production node making a few plain-HTTP/1.1 JSON-RPC calls per second.

**Before** — `hackney_conn` process count climbing ~1.5/s (1714 → 2539) while only ~34 OS ports exist; every process `:sys.get_state` → `connected`, `checkin_info/1` → `no_reuse => true`, sole link is `hackney_conn_sup`. Socketless zombies.

**After** — sampling the process count every 15s:

```
conns=0  ports=32
conns=4  ports=36
conns=6  ports=38
conns=0  ports=32
conns=0  ports=32
conns=0  ports=32
conns=5  ports=37
conns=5  ports=37
conns=0  ports=32
conns=0  ports=32
conns=1  ports=33
conns=2  ports=34
```

The count oscillates near zero, **returns to 0** between request bursts, and tracks live OS ports 1:1 (`ports = 32 baseline + conns`) — every surviving process is a genuine in-flight request, not an orphan. Spot-checking survivor states returned `%{connected: 1}` (one in-flight request) vs `%{connected: <thousands>}` before.

Verified against `master` (also reproduces on the 4.5.2 release; the sync path is identical). Module compiles clean.

### DIFF
--- a/src/hackney_conn.erl
+++ b/src/hackney_conn.erl
@@ -1518,8 +1518,9 @@ receiving({call, From}, body, Data) ->
             %% Transition to closed state instead of connected
             {next_state, closed, NewData, [{reply, From, {ok, Body}}]};
         {ok, Body, NewData} ->
-            %% Socket still valid - return to connected state
-            {next_state, connected, NewData, [{reply, From, {ok, Body}}]};
+            %% Socket still valid - reuse it, or stop the process if the
+            %% connection must not be reused (see finish_sync_request/3).
+            finish_sync_request(From, {ok, Body}, NewData);
         {error, Reason} ->
             {next_state, closed, Data, [{reply, From, {error, Reason}}]}
     end;
@@ -1533,7 +1534,7 @@ receiving({call, From}, stream_body, Data) ->
             %% Socket was closed during body read - transition to closed state
             {next_state, closed, NewData, [{reply, From, done}]};
         {done, NewData} ->
-            {next_state, connected, NewData, [{reply, From, done}]};
+            finish_sync_request(From, done, NewData);
         {error, Reason} ->
             {next_state, closed, Data, [{reply, From, {error, Reason}}]}
     end;
@@ -1930,6 +1931,29 @@ should_close_connection(#conn_data{response_headers = undefined, version = undef
 should_close_connection(#conn_data{version = Version, response_headers = Headers,
                                    request_close = RequestClose}) ->
     hackney_keepalive:should_close(Version, Headers, RequestClose).
+
+%% @private Finish a completed synchronous request. The blocking `body` /
+%% `stream_body` path historically returned unconditionally to `connected`,
+%% never consulting `no_reuse` or `should_close_connection/1`. A connection that
+%% must not be reused — flagged `no_reuse` (proxy tunnels, SSL upgrades, pool
+%% disabled) or carrying `Connection: close` per RFC 7230 — therefore parked in
+%% `connected` forever, pinned alive by its owner link, leaking one hackney_conn
+%% process per request. This mirrors finish_async_streaming/1 for the sync path:
+%% close the socket and stop the process for non-reusable connections, otherwise
+%% return to `connected` for keep-alive reuse.
+finish_sync_request(From, Reply, #conn_data{transport = Transport, socket = Socket,
+                                            no_reuse = NoReuse} = Data) ->
+    case NoReuse orelse should_close_connection(Data) of
+        true ->
+            case Socket of
+                undefined -> ok;
+                _ -> Transport:close(Socket)
+            end,
+            {stop_and_reply, normal, [{reply, From, Reply}],
+             Data#conn_data{socket = undefined}};
+        false ->
+            {next_state, connected, Data, [{reply, From, Reply}]}
+    end.
 
 %% @private Finish async streaming - close or return to connected based on Connection header
 finish_async_streaming(Data) ->


benoitc:
Thanks for the detailed diagnosis. Superseded by #908, which unifies the sync and async reuse decision into a single shared check.

Issue #895, opened 2026-07-03
HTTP/2 request bodies over 1 MB abort with {error, send_buffer_full}


barttenbrinke:
## Environment
* hackney 4.4.5, h2 0.10.2, Erlang/OTP 28.4, Elixir 1.19.5 (also seen via Tesla.Adapter.Hackney and HTTPoison 3.0.0, but the behavior is in hackney/h2)
* Peer: a standard HTTP/2 server over TLS (ALPN negotiates h2)

Since upgrading 1.x → 4.x, any POST/PUT with a body larger than ~1 MB over HTTP/2 intermittently fails with {error, send_buffer_full}. Same requests worked over HTTP/1.1 (hackney 1.x) and work again in 4.x with {protocols, [http1]}. Reproduces on real uploads: multipart documents and XML/JSON with base64 attachments (~1–5 MB).

## Possible cause
h2 caps a stream's send buffer at ?MAX_SEND_BUFFER_BYTES (1 MB) and returns {error, send_buffer_full} as intentional backpressure ([h2_connection.erl:69 + :2577]). hackney streams the body with the non-blocking send and propagates that error, aborting the upload - it never waits for WINDOW_UPDATEs to drain, nor uses h2's blocking h2:send_data/5 (#{block => Timeout}) variant ([hackney_conn.erl stream_body_fun_h2, ~:3020]).

So whenever hackney feeds chunks faster than the peer opens its window (normal for a multi-MB body against the default 64 KB initial window), the buffer crosses 1 MB and the request dies.

## Suggested fix
I think h2 already provides the primitive: `h2_connection:send_data/5` with `#{block => Timeout}` "parks the caller until the buffer drains, returning `ok`, or `{error, timeout}`." - https://github.com/benoitc/erlang_h2/blob/0.10.2/README.md#L253  Both hackney body-send paths (one-shot and `stream_body_fun_h2`) funnel through the single `h2_send_data/4` helper, so switching that one call to the blocking form should fix all of them:

```erlang
%% deps/hackney/src/hackney_conn.erl
 h2_send_data(H2Conn, StreamId, Bin, EndStream) ->
     try
-        h2_connection:send_data(H2Conn, StreamId, Bin, EndStream)
+        h2_connection:send_data(H2Conn, StreamId, Bin, EndStream, #{block => SendTimeout})
     catch
         exit:{ExitReason, _} -> {error, {closed, ExitReason}};
         exit:ExitReason      -> {error, {closed, ExitReason}}
[...]

PR #905, opened 2026-07-13
fix: HTTP/2 sends now wait up to 5 seconds for flow-control credit instead of returning send_buffer_full

speeddragon:
Disclaimer: This PR code was generated by AI.

I'm trying to address #895. I'm having the same issue with `ex_aws`, and this PR modifies the request to wait until a specific time rather than returning `{error, send_buffer_full}`.

This change allows waiting for the buffer to clear instead of returning an error.

How to use it with ExAws:
```elixir
  config :ex_aws, :hackney_opts,
    h2_send_timeout: 5_000
```
  Or per request:
```elixir
  ExAws.request(operation, http_opts: [h2_send_timeout: 10_000]) 
```

## TODO
- [x] Run `rebar3 eunit`
- [ ] Update documentation about the new property

### DIFF
--- a/src/hackney.erl
+++ b/src/hackney.erl
@@ -137,6 +137,7 @@ connect_direct(Transport, Host, Port, Options) ->
     transport => Transport,
     connect_timeout => proplists:get_value(connect_timeout, Options, 8000),
     recv_timeout => proplists:get_value(recv_timeout, Options, 5000),
+    h2_send_timeout => proplists:get_value(h2_send_timeout, Options, 5000),
     connect_options => ConnectOpts,
     ssl_options => proplists:get_value(ssl_options, Options, [])
   },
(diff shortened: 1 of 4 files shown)


benoitc:
It seems that what you want is  blocking send i but is there any reason you need a configurable send wait per request ?  I don't see a real use case for `h2_send_timeout`and the complexity it adds.

This will change the behabiour though, instead of returnong once the 1MB buffer cap is crossed `{error, send_buffer_full}` , the application will have to wait for a `{error, timeout}`

speeddragon:
Thanks for the feedback.

> It seems that what you want is blocking send i but is there any reason you need a configurable send wait per request ?

It doesn't need to be per request. I'm not sure we can define a more low-level place and have it picked up by other high-level (connection vs request).

I've also noticed this doesn't preserve the default behaviour. I will make this optional.

> This will change the behabiour though, instead of returnong once the 1MB buffer cap is crossed `{error, send_buffer_full}` , the application will have to wait for a `{error, timeout}`

Is there a way to handle `send_buffer_full` outside of hackney? Retry doesn't look like an option because will lead to the same error. Modifying the buffer size doesn't look great as well. 

Maybe I'm missing the right way to handle send_buffer_full outside of hackney.

benoitc:
Thanks, I finally fixed it in #911, released in 4.7.0. Main differences: blocking is the default now (`send_timeout` option, 30s, `nonblock` to opt out), the stream is reset on timeout so its buffer does not stay on a shared connection, and the per request timeout does not leak between pooled requests. Closing.

PR #917, opened 2026-07-23
Always add Content-Length: 0 even for empty bodies

onno-vos-dev:
I got a report on https://github.com/aws-beam/aws-elixir/issues/238 that the upgrade I recently did to a later version of Hackney broke things for 0-sized bodies. AWS apparently requires a Content-Length being set even for 0-sized bodies.

Instead of implementing a fix on my end, I'm testing the waters to see how open you are to fixing this from within Hackney itself.

### DIFF
--- a/src/hackney_conn.erl
+++ b/src/hackney_conn.erl
@@ -2123,8 +2123,12 @@ build_headers(_Method, Headers0, Body, Netloc) ->
 
     %% Add Content-Length for bodies
     case Body of
-        <<>> -> Headers3;
-        [] -> Headers3;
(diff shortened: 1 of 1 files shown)


benoitc:
Thanks @onno-vos-dev for the report and for flagging the AWS regression (aws-beam/aws-elixir#238).

I picked this up in #925, done the curl way: an empty body gets Content-Length: 0 only for the body-bearing methods (POST/PUT/PATCH), and using a binary value so it stays type-correct. GET/HEAD/DELETE stay without it.

Closing in favor of #925, which covers your case.

PR #923, opened 2026-08-09
Drop request body on cross-origin 307/308 redirects

benoitc:
A 307/308 keeps the method and body, but the body was forwarded to the redirect target regardless of origin; only auth and cookie options were stripped. A POST/PUT/PATCH body still reached a different host and leaked request contents, across the HTTP/1.1, HTTP/2 and HTTP/3 paths. Clear the body when the redirect leaves the original host unless the caller set `location_trusted`, matching the credential-header stripping.

### DIFF
--- a/src/hackney.erl
+++ b/src/hackney.erl
@@ -1477,7 +1477,14 @@ follow_redirect(ConnPid, Method, Body, WithBody, Options, CurrentURL, RespHeader
       end,
       NewBody = case NewMethod of
         <<"GET">> -> <<>>;
-        _ -> Body
+        _ ->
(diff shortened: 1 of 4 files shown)


benoitc:
After checking against RFC 9110, I close this one.

For a 307/308 the client must keep the same method and the body, this is exactly why these codes exist. And the browsers, node-fetch, curl, they all keep the body on a cross-origin redirect too, they only strip the credential headers (Authorization, Cookie, Proxy-Authorization), which we already do. So dropping the body is not correct here, it would break the legit cases like the S3 presigned upload or an API gateway hand-off.

The real risk on cross-origin is the credentials, and this part is already handled. So no need for this change.

Issue #927, opened 2026-08-11
4.x regression, a slow connect crashes the caller


jegaxd26:
Hi we're in the middle of upgrading to hackney 4.x and stumbled upon this regression (presumably) when a client taking more then 8 seconds to accept the connection will crash the caller. I double checked that this is not the case with 1.x

### Environment

- hackney 4.7.2
- Erlang/OTP 27 (erts-15.2.6)
- Elixir 1.18.3
- OS X

a script to reproduce:

```elixir
Mix.install([{:hackney, "~> 4.0"}])

Application.ensure_all_started(:hackney)

defmodule StalledHackneyTransport do

  def connect(_host, _port, _options, _timeout) do
    # hackney_conn:connect/1 has a fixed 8-second call timeout. Ignoring the
    # transport timeout reproduces a stalled connect that outlives that call.
    Process.sleep(8_500)
    {:error, :timeout}
  end
end

defmodule Test do
  @pool_name :hackney_connect_timeout_regression
  
  def run do
    :ok = :hackney_pool.start_pool(@pool_name, [])

    pool_pid = :hackney_pool.find_pool(@pool_name)

    monitor_ref = Process.monitor(pool_pid)

    {:error, _reason} =
      :hackney_pool.checkout(
        ~c"stalled.example",
        443,
        StalledHackneyTransport,
        pool: @pool_name,
        connect_timeout: 100,
        checkout_timeout: 100
      )

    receive do
      {:DOWN, ^monitor_ref, :process, ^pool_pid, _reason} ->
        IO.puts "Bug"
    after 9000 ->
        IO.puts "No bug"
    end
  end
end

Test.run
```

when run, with 4.0 it prints "Bug" and throws this exception:

```
[error] GenServer #PID<0.208.0> terminating
** (stop) exited in: :gen_statem.call(#PID<0.209.0>, :connect, 8000)
    ** (EXIT) time out
    (stdlib 6.2.2) gen.erl:277: :gen.do_call/4
    (stdlib 6.2.2) gen_statem.erl:3250: :gen_statem.call/3
    (hackney 4.7.2) /Users/Library/Caches/mix/installs/elixir-1.18.3-erts-15.2.6/d8d5a2166b8316f756e16fc57188eab6/deps/hackney/src/hackney_pool.erl:1017: :hackney_pool.start_connection/6
[...]

PR #929, opened 2026-08-11
Fault injection and chaos tests for the pool

benoitc:
Based on #928, so it contains its two commits. Merge #928 first, a rebase then removes them.

Our tests only use servers which answer. Nothing covered a connection which stalls, crashes, or dies during a checkout, the case of #927.

New modules: `hackney_fault_transport`, a transport identical to `hackney_tcp` until a fault is armed on one callback; `hackney_crash_sentinel`, which makes a dead pool visible under `error_logger:tty(false)`; fault tests for the checkout entry points and for the shared HTTP/2 and HTTP/3 connections; a chaos run, longer with `HACKNEY_CHAOS_WORKERS` and `HACKNEY_CHAOS_ROUNDS`; and `hackney_pool_safety_tests`, which reads the compiled code and fails if the pool calls a connection process outside a `try`.

Three problems found, corrected here:

1. Four unprotected calls of the same kind as #927: three `hackney_conn:set_owner/2` and the prewarm dial.
2. `stop_conn` used `gen_statem:stop/1`, which waits without limit, so after a dial timeout the pool stayed blocked for the whole duration of the transport call. It is now bounded to 100ms, then kills the connection.
3. `h2_conn_usable/1` used `get_state/1`, which waits 5s, so one blocked HTTP/2 connection blocked the pool 5s per checkout on that host. `is_ready`, `checkin_info` and `set_owner` had the same problem. The four functions now take a timeout and the pool passes 250ms.

eunit is green (1096 tests), xref and dialyzer are clean. `DEVELOPMENT.md` explains how to arm a fault and how to run a longer chaos test.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -66,6 +66,57 @@ gunicorn -b 127.0.0.1:8000 httpbin:app &
 rebar3 eunit
 ```
 
+### Fault injection and chaos tests
+
+Use these when you change the pool or the connection state machine. Ordinary
+integration tests only exercise servers that answer, so they never see a
+connection that stalls, crashes, or dies at the wrong moment, which is where
+pool failures come from: the pool dials and stops connections from inside its
+own gen_server, and an unguarded call that exits takes the pool down along with
+every caller using it.
+
+Three pieces make up the harness:
+
+| Module | What it does |
+| --- | --- |
+| `hackney_fault_transport` | A transport that behaves like `hackney_tcp` until you arm a fault on one of its callbacks |
+| `hackney_crash_sentinel` | Captures crash reports so a test can assert a process survived, even with `error_logger:tty(false)` |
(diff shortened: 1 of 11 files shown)


benoitc:
Extended to the multiplexed checkout paths in 721cf78.

`h2_conn_usable/1` carried a comment claiming a short timeout but called `hackney_conn:get_state/1`, which waits the 5s default, so a single wedged HTTP/2 connection stalled the pool for 5s on every checkout for that host. `is_ready`, `checkin_info` and `set_owner` had the same exposure. All four now take an explicit timeout and the pool passes 250ms.

`hackney_pool_h2h3_fault_tests` wedges a registered connection with `sys:suspend/1` and fails if the pool waits on it. Before the change that scenario hit the eunit timeout; it now answers `none` in 252ms.

Issue #914, opened 2026-07-20
connect_pool: unguarded hackney_conn:get_state/1 crashes checkout when a pooled HTTP/2/3 connection terminates mid-call


kivra-kenand:
### Environment
  - hackney: 4.7.2 (also present in 4.5.x)
  - Erlang/OTP: 28
  - Observed in production under HTTP/2 connection reuse to a remote that idle-closes / GOAWAYs pooled connections.
  
  ### Summary
  When a pooled HTTP/2 (or HTTP/3) connection is being reused, `connect_pool/4` validates it with `hackney_conn:get_state/1`, which is a bare `gen_statem:call/2`. If that pooled connection process is terminating at the exact moment of checkout (idle teardown, server GOAWAY, keepalive close), the `gen_statem:call` **exits**. The surrounding `case` only handles return values, so the exit propagates and crashes the caller of `hackney:connect/4` instead of falling back to a fresh connection.

  ### Affected code
  `src/hackney.erl`, HTTP/2 checkout in `connect_pool/4` (~L200):
        
      case PoolHandler:checkout_h2(Host, Port, Transport, Options2) of
        {ok, H2Pid} ->
          case hackney_conn:get_state(H2Pid) of      %% <-- exits if H2Pid is terminating
            {ok, connected} -> {ok, H2Pid};
            _ -> PoolHandler:unregister_h2(H2Pid, Options2),
                 connect_pool_new(...)
          end;
        none -> connect_pool_new(...)
      end

  The same unguarded pattern exists for HTTP/3 in `try_h3_connection/5` (~L251), using `hackney_conn:get_state(H3Pid)`.

  `hackney_conn:get_state/1` (src/hackney_conn.erl):

      get_state(Pid) -> gen_statem:call(Pid, get_state).

  ### Why this looks like an oversight
  The same file already defends against this exact race for the sibling calls. `maybe_register_h2/6` and `maybe_upgrade_ssl/3` wrap their `hackney_conn:*` calls with the comment:

      %% Wrapped in try to handle a race where the connection terminates before the call

  The two `get_state` checkout calls just weren't given the same treatment.

  ### Symptom (caller-side)
  `hackney:connect/4` exits (rather than returning `{error, _}`) with a reason of the form:
      
      {noproc,  {gen_statem, call, [<Pid>, get_state | _]}}
      {normal,  {gen_statem, call, [<Pid>, get_state | _]}}
      {shutdown,{gen_statem, call, [<Pid>, get_state | _]}}

[...]

Issue #850, opened 2026-05-19
Intermittent Pool GenServer crash on `hackney_conn:set_owner/2`  failing to match `{error, invalid_state}` if the connection is closed


ashutoshrishi:
> **Note:** This issue was debugged with the assistance of Claude. I
> have verified the analysis to the best of my knowledge.

**Version:** 4.0.0

## What happens

The hackney pool GenServer crashes with a `MatchError` when a checkout races a
connection closing. `hackney_sup` restarts the pool immediately (`one_for_one`,
`permanent`) and `checkout/4`'s `try/catch` converts the exit into `{error, ...}`
for the caller, so the system recovers. **The practical impact is a single failed
checkout and log noise.**

Noticed this when using
[:broadway_sqs](https://github.com/dashbitco/broadway_sqs), which polls SQS
regularly. AWS closes idle connections on their end, triggering the race. 

Have had similar experiences with other HTTP clients where AWS closing
connections needs handling in application or client library level.

```
** (MatchError) no match of right hand side value: {:error, :invalid_state}
    (hackney 4.0.0) hackney_pool.erl:450: :hackney_pool.handle_call/3
    (stdlib 7.3) gen_server.erl:2470: :gen_server.try_handle_call/4
    (stdlib 7.3) gen_server.erl:2499: :gen_server.handle_msg/3
    (stdlib 7.3) proc_lib.erl:333: :proc_lib.init_p_do_apply/3
Last message: {:checkout, {"sqs.eu-west-2.amazonaws.com", 443, :hackney_tcp}, #PID<...>, [recv_timeout: 30000]}
```

## Suspected root cause

`hackney_pool.erl:450` hard-matches the return of `hackney_conn:set_owner/2`:

```erlang
ok = hackney_conn:set_owner(Pid, Requester),
```

`find_available` does call `hackney_conn:is_ready/1` and only proceeds when it
gets `{ok, connected}`. It skips a connection in `closed` state.

Maybe: `is_ready` and `set_owner` are two separate synchronous
`gen_statem` calls, which may leave a window where a `tcp_closed` info message
arrives in the connection process mailbox between them:

1. Pool calls `is_ready`. gen_statem responds `{ok, connected}`, mailbox empty
2. TCP stack delivers `tcp_closed` to the gen_statem mailbox because AWS closed the connection?
3. Pool receives `{ok, connected}` and dispatches `set_owner`
4. gen_statem processes `tcp_closed` first (queued earlier). Transitions to `closed`
[...]

Issue #544, opened 2018-11-07
Reused socket closing connection


gutiago:
Hello!
I am having some problems related to the sockets being reused in my pool. For some reason, it's returning `{error, closed}` when hackney calls `hackney_tcp:recv(Socket, Length, Timeout)`.

To illustrate the situation, I've created some test cases for my remote server. I perform some requests using hackney and after 5 seconds I start it again, so it's in a loop. In the low test, I make 3 requests per loop.

But I've noticed that in some cases, the request was returning {error, closed}. I took a look at the methods that my request should call and added some prints to the code.

I decided to print when the socket was connected and also when became reused. To have a better idea which socket was disconnecting, I decided to print inside the `hackney_tcp: recv` the info of the socket. So the code looks like:

```
recv(Socket, Length, Timeout) ->
        SocketInfo = {sockname(Socket), peername(Socket)},
        io:format("RCV ~p~n", [SocketInfo]),
        case gen_tcp:recv(Socket, Length, Timeout) of
           {error, closed} ->
                io:format("ERROR ~p~n", [SocketInfo]),
                {error, closed};
           R -> R
        end.
```

After executing the test, I got the following logs: 

> note that as I call 3 requests per iteration, hackney create (CONN) / reuse a connection and call rcv.
> I log the operation with the info of the socket  {sockname(Socket), peername(Socket)}

**First Iteration:** 

> CONN {{ok,{{10,36,20,25},64003}}, {ok,{{10,39,252,202},8082}}}
> RCV {{ok,{{10,36,20,25},64003}}, {ok,{{10,39,252,202},8082}}}
> CONN {{ok,{{10,36,20,25},14409}},{ok,{{10,39,252,202},8082}}}
> RCV {{ok,{{10,36,20,25},14409}},{ok,{{10,39,252,202},8082}}}
> CONN {{ok,{{10,36,20,25},22965}},{ok,{{10,39,252,202},8082}}}
> RCV {{ok,{{10,36,20,25},22965}},{ok,{{10,39,252,202},8082}}}
> 

**Second Iteration:**

> CONN {{ok,{{10,36,20,25},39219}},{ok,{{10,39,252,202},8082}}}
> RCV {{ok,{{10,36,20,25},39219}},{ok,{{10,39,252,202},8082}}}
> REUSE {{ok,{{10,36,20,25},22965}},{ok,{{10,39,252,202},8082}}}
[...]

PR #934, opened 2026-08-28
fix: cancel the keepalive timer when a connection is checked out

lennartschoch:
Ran into this after moving an Elixir app onto hackney 4. We started getting intermittent `{error, invalid_state}` back from Stripe calls, a handful an hour, with nothing else looking wrong.

Turns out `connected(enter)` arms a `keepalive_timeout` `state_timeout` on pooled conns and nothing on the checkout path resets it. `is_ready/1` does a decent job on the socket itself (`has_pending_close` for #544, then `check_socket_health`) but it leaves the timer alone, so a conn can pass the probe, get handed over, and then close itself because the timer had been running since well before the probe. The request turns up after that and falls through `handle_common` as `invalid_state`.

The fix just disarms the timer in the `is_ready` branch that returns `{ok, connected}`, since a conn heading out to a requester isn't idle any more. `connected(enter)` re-arms it when the conn comes back to the pool, so one that gets abandoned still ages out. It's the same `{state_timeout, infinity, idle_timeout}` you already do on the h2 paths, which is why we only ever saw this on http1.

Probably why it hasn't come up much: you need a slow caller for the window to matter (odds are about the checkout-to-write gap over `keepalive_timeout`, and a pooled conn is usually near its deadline anyway), and it can't happen on h2 at all. Feels like the same family as #850, #869 and #914, just on the http1 side.

`rebar3 eunit` is green, 1099 tests. Added `test/hackney_pool_checkout_idle_tests.erl` — two of those fail without the patch, one seeing the conn already `closed` after a good probe and the other getting `invalid_state` from the request. The third is a control that the timer is still armed on release, and it passes either way.


benoitc:
Thanks, merging.

